PR #2, opened 2020-10-02
Ports over most of the tests from Contracts V1

kfichter:
This PR is really big and it's hard to describe without just looking at the diff.


ben-chain (approved):
Woot woot LGTM!!

PR #3, opened 2020-10-05
Bond Manager

gakonst:
Enables submitting fidelity bonds by the sequencers which they forfeit in cause of fraud. The reward gets distributed proportionally to all the users that provided witness for the dispute resolution.

### DIFF
--- a/contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
+++ b/contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
@@ -51,7 +51,7 @@ contract OVM_BondManager is Lib_AddressResolver {
     mapping(address => Bond) public bonds;
 
     /// The dispute period
-    uint256 public forceInclusionPeriodSeconds = 7 days;
+    uint256 public disputePeriodSeconds = 7 days;
 
     // Per pre-state root, store the number of state provisions that were made
     // and how many of these calls were made by each user. Payouts will then be
@@ -110,7 +110,7 @@ contract OVM_BondManager is Lib_AddressResolver {
         // block's challenge period
         if (
             bond.withdrawalTimestamp != 0 && 
-            bond.withdrawalTimestamp <= timestamp + forceInclusionPeriodSeconds &&
+            bond.withdrawalTimestamp <= timestamp + disputePeriodSeconds &&
             bond.withdrawing >= requiredCollateral
         ) {
             bond.withdrawing -= requiredCollateral;
@@ -133,7 +133,7 @@ contract OVM_BondManager is Lib_AddressResolver {
         Bond storage bond = bonds[msg.sender];
 
         require(
-            block.timestamp >= bond.withdrawalTimestamp + forceInclusionPeriodSeconds, 
+            block.timestamp >= bond.withdrawalTimestamp + disputePeriodSeconds, 
             Errors.TOO_EARLY
         );
         require(bond.withdrawing >= requiredCollateral, Errors.SLASHED);
--- a/contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
+++ b/contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
@@ -168,8 +168,8 @@ contract OVM_BondManager is Lib_AddressResolver {
     // Collateral Management
     ////////////////////////
 
-    // Stakes the user for the provided batch index
-    function stake(address who, uint256 batchIndex) public view returns (bool) {
+    // Checks if the user is collateralized for the batchIndex
+    function isCollateralized(address who, uint256 batchIndex) public view returns (bool) {
         require(msg.sender == ovmCanonicalStateCommitmentChain, Errors.ONLY_STATE_COMMITMENT_CHAIN);
         require(bonds[who].locked >= requiredCollateral, Errors.NOT_ENOUGH_COLLATERAL);
         return true;
--- a/contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
+++ b/contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
@@ -12,11 +12,11 @@ interface ERC20 {
 /// All the errors which may be encountered on the bond manager
 library Errors {
     string constant ERC20_ERR = "BondManager: Could not post bond";
-    string constant NOT_ENOUGH_COLLATERAL = "BondManager: Sequencer is not sufficiently collateralized";
     string constant LOW_VALUE = "BondManager: New collateral value must be greater than the previous one";
     string constant HIGH_VALUE = "BondManager: New collateral value cannot be more than 5x of the previous one";
     string constant ALREADY_FINALIZED = "BondManager: Fraud proof for this pre-state root has already been finalized";
     string constant SLASHED = "BondManager: Cannot finalize withdrawal, you probably got slashed";
+    string constant WRONG_STATE = "BondManager: Wrong bond state for proposer";
     string constant CANNOT_CLAIM = "BondManager: Cannot claim yet. Dispute must be finalized first";
 
     string constant WITHDRAWAL_PENDING = "BondManager: Withdrawal already pending";
@@ -34,14 +34,22 @@ contract OVM_BondManager is Lib_AddressResolver {
      * Data Structures *
      *******************/
 
+    /// The lifecycle of a proposer's bond
+    enum State {
+        // Before depositing or after getting slashed, a user is uncollateralized
+        NOT_COLLATERALIZED,
+        // After depositing, a user is collateralized
+        COLLATERALIZED,
+        // After a user has initiated a withdrawal
+        WITHDRAWING
+    }
+
     /// A bond posted by a proposer
     struct Bond {
-        // The amount a proposer has posted as collateral
-        uint256 locked;
-        // The amount a proposer is currently withdrawing
-        uint256 withdrawing;
+        // The user's state
+        State state;
         // The timestamp at which a proposer issued their withdrawal request
-        uint256 withdrawalTimestamp;
+        uint32 withdrawalTimestamp;
     }
 
     // Per pre-state root, store the number of state provisions that were made
@@ -134,20 +142,22 @@ contract OVM_BondManager is Lib_AddressResolver {
 
         Bond storage bond = bonds[publisher];
 
-        // always slash 1 round of collateral, if possible
-        if (bond.locked >= requiredCollateral) {
-            bond.locked -= requiredCollateral;
-        }
-
-        // if the publisher has a pending withdrawal that's within the published
-        // block's challenge period
+        // if the fraud proof's dispute period does not intersect with the 
+        // withdrawal's timestamp, then the user should not be slashed
+        // e.g if a user at day 10 submits a withdrawal, and a fraud proof
+        // from day 1 gets published, the user won't be slashed since day 8 (1d + 7d)
+        // is before the user started their withdrawal. on the contrary, if the user
+        // had started their withdrawal at, say, day 6, they would be slashed
         if (
             bond.withdrawalTimestamp != 0 && 
-            bond.withdrawalTimestamp <= timestamp + disputePeriodSeconds &&
-            bond.withdrawing >= requiredCollateral
+            uint256(bond.withdrawalTimestamp) > timestamp + disputePeriodSeconds &&
+            bond.state == State.WITHDRAWING
         ) {
-            bond.withdrawing -= requiredCollateral;
+            return;
         }
+
+        // slash!
+        bond.state = State.NOT_COLLATERALIZED;
     }
 
     /// Sequencers call this function to post collateral which will be used for
@@ -159,18 +169,17 @@ contract OVM_BondManager is Lib_AddressResolver {
         );
 
         // This cannot overflow
-        bonds[msg.sender].locked += amount;
+        bonds[msg.sender].state = State.COLLATERALIZED;
     }
 
     /// Starts the withdrawal for a publisher
     function startWithdrawal() public {
         Bond storage bond = bonds[msg.sender];
         require(bond.withdrawalTimestamp == 0, Errors.WITHDRAWAL_PENDING);
-        require(bond.locked >= requiredCollateral, Errors.NOT_ENOUGH_COLLATERAL);
-        bond.locked -= requiredCollateral;
+        require(bond.state == State.COLLATERALIZED, Errors.WRONG_STATE);
 
-        bond.withdrawing += requiredCollateral;
-        bond.withdrawalTimestamp = block.timestamp;
+        bond.state = State.WITHDRAWING;
+        bond.withdrawalTimestamp = uint32(block.timestamp);
     }
 
     /// Finalizes a pending withdrawal from a publisher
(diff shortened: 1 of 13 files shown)


contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,176 @@
...
+        bonds[msg.sender] -= amount;
+        require(
+            token.transfer(msg.sender, amount),
+            Errors.ERC20_ERR
+        );
+    }
+
+    /// Sets the required collateral for posting a state root
+    /// Callable only by the contract's deployer.
+    function setRequiredCollatreal(uint256 newValue) public {
```
MrChico, line 167: typo :)

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,176 @@
...
+        require(newValue > requiredCollateral, Errors.LOW_VALUE);
```
kfichter, line 168: Should we limit this by some % increase per update?

contracts/optimistic-ethereum/OVM/verification/OVM_FraudVerifier.sol
```diff
@@ -178,13 +179,25 @@ contract OVM_FraudVerifier is iOVM_FraudVerifier, Lib_AddressResolver {
...
         );
 
-        require(
-            _postStateRoot != transitioner.getPostStateRoot(),
-            "State transition has not been proven fraudulent."
-        );
+        // If the post state root did not match, then there was fraud and we should
+        // delete the batch
+        bool isFraud;
+        if (_postStateRoot != transitioner.getPostStateRoot()) {
```
kfichter, line 185: I don't think we need to do this check -- if the transaction wasn't fraudulent then we shouldn't need to punish anyone or reward the users who contributed to the attempted fraud proof. Essentially the same as my comment above within `OVM_BondManager.finalize`.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -13,6 +13,7 @@ import { iOVM_StateTransitioner } from "../../iOVM/verification/iOVM_StateTransi
 import { iOVM_ExecutionManager } from "../../iOVM/execution/iOVM_ExecutionManager.sol";
```
kfichter, line 13: I think we also need to give some of the reward to the user who calls `executeTransaction`.

kfichter (commented):
Good first stab at this, I think there are a few changes/simplifications we can make (especially around `OVM_BondManager.finalize`). But generally I think this flow is correct.

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    /// The fraud verifier contract, used to get data about transitioners for a pre-state root
+    // address public ovmFraudVerifier;
+    address public ovmCanonicalStateCommitmentChain;
+
+    uint256 requiredCollateral = 1 ether;
+
+    /// A bond posted by a sequencer
+    /// The bonds posted by each sequencer
+    mapping(address => uint256) public bonds;
```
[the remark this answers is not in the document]
gakonst: That's helpful, I didn't realize that.

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,176 @@
...
+        witnessProviders[_preStateRoot].numClaims[who] += 1;
+    }
+
+    /// Slashes + distributes rewards or frees up the sequencer's bond, only called by
+    /// `FraudVerifier.finalizeFraudVerification`
+    function finalize(bytes32 _preStateRoot, uint256 batchIndex, bool isFraud) public {
+        require(msg.sender == resolve("OVM_FraudVerifier"), "ERR not callable by non fraud verifier");
+
+        // TODO: can this be removed?
+        require(sequencers[batchIndex] != address(0), "err: sequencer already claimed");
```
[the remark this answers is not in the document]
gakonst: Yeah so IMO let's keep it as-is, and remove it once we have..._finalized_ the design and have confidence in `OVM_FraudVerifier`

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
gakonst: Yeah sure, whatever you think is an appropriate % value. @ben-chain?

contracts/optimistic-ethereum/OVM/verification/OVM_FraudVerifier.sol
gakonst: Reverted back to the `require` statement

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,176 @@
...
+        if (isFraud) {
+            // allow users to claim from that state root's
+            // pool of collateral (effectively slashing the sequencer)
+            witnessProviders[_preStateRoot].canClaim = true;
+        } else {
+            // refund collateral to the sequencer for that batch
+            // TODO: Should this actually be refunded? Can't there be more than 1
+            // cases of fraud for a batch e.g. because of multiple invalid state
+            // transitions? should these be penalized multiple times?
+            bonds[sequencers[batchIndex]] += requiredCollateral;
```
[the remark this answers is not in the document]
gakonst: Updated

contracts/optimistic-ethereum/OVM/chain/OVM_StateCommitmentChain.sol
```diff
@@ -58,6 +62,12 @@ contract OVM_StateCommitmentChain is iOVM_StateCommitmentChain, OVM_BaseChain, L
         override
         public
     {
+        // Stake!
+        require(
+            ovmBondManager.stake(msg.sender, batches.length + 1),
```
gakonst, line 67: TODO: Rename `stake` to `isCollateralized`

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,201 @@
...
+    struct Bond {
+        uint256 locked;
+        uint256 withdrawing;
+        uint256 withdrawalTimestamp;
+
+    }
+    mapping(address => Bond) public bonds;
+
+    /// The dispute period
+    uint256 public forceInclusionPeriodSeconds = 7 days;
```
kfichter, line 54: Totally our fault here -- `forceInclusionPeriodSeconds` is a variable in the `CanonicalTransactionChain` that should be renamed. It's meant to designate the period of time in which the sequencer can add transactions in advance of the `L1ToL2TransactionQueue`. We should probably change that variable inside the CTC to be more clear. 

Will want to rename `forceInclusionPeriodSeconds` here to `disputePeriodSeconds` or similar.

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        // if the publisher has a pending withdrawal that's within the published
+        // block's challenge period
+        if (
+            bond.withdrawalTimestamp != 0 && 
+            bond.withdrawalTimestamp <= timestamp + forceInclusionPeriodSeconds &&
+            bond.withdrawing >= requiredCollateral
+        ) {
+            bond.withdrawing -= requiredCollateral;
```
kfichter, line 116: I think a simpler method here would be to set `bond.locked = 0` (no need to reduce exactly by `requiredCollateral`) and then assert that `bond.withdrawing <= bond.locked` within `finalizeWithdrawal` and that `bond.withdrawing - bond.locked >= requiredCollateral` for within `stake`

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+
+    /// Starts the withdrawal for a publisher
+    function startWithdrawal() public {
+        Bond storage bond = bonds[msg.sender];
+        require(bond.withdrawalTimestamp == 0, Errors.WITHDRAWAL_PENDING);
+        require(bond.locked >= requiredCollateral, Errors.NOT_ENOUGH_COLLATERAL);
+        bond.locked -= requiredCollateral;
```
kfichter, line 125: Yeah with above method (checking `bond.withdrawing <= bond.locked`) we can avoid changing `bond.locked` within this function.

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,201 @@
...
+        // transfer
+        require(token.transfer(msg.sender, amount), Errors.ERC20_ERR);
+    }
+
+    ////////////////////////
+    // Collateral Management
+    ////////////////////////
+
+    // Stakes the user for the provided batch index
+    function stake(address who, uint256 batchIndex) public view returns (bool) {
```
kfichter, line 172: Reminder on renaming to `isCollateralized`.

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,201 @@
...
+    mapping(bytes32 => Rewards) public witnessProviders;
+
+    /// Initializes with a ERC20 token to be used for the fidelity bonds
+    /// and with the Address Manager
+    constructor(ERC20 _token, address _libAddressManager)
+        Lib_AddressResolver(_libAddressManager)
+    {
+        owner = msg.sender;
+        token = _token;
+        // ovmFraudVerifier = resolve("OVM_FraudVerifier"); // TODO: Re-enable this
```
kfichter, line 78: What is this pending? Just to confirm the meaning of the TODO.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -70,13 +72,15 @@ contract OVM_StateTransitioner is iOVM_StateTransitioner, Lib_AddressResolver {
     )
         Lib_AddressResolver(_libAddressManager)
     {
+        // TODO: Can this be removed?
```
kfichter, line 75: Can what be removed?

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -70,13 +72,15 @@ contract OVM_StateTransitioner is iOVM_StateTransitioner, Lib_AddressResolver {
...
         stateTransitionIndex = _stateTransitionIndex;
         preStateRoot = _preStateRoot;
         postStateRoot = _preStateRoot;
         transactionHash = _transactionHash;
 
         ovmExecutionManager = iOVM_ExecutionManager(resolve("OVM_ExecutionManager"));
         ovmStateManager = iOVM_StateManagerFactory(resolve("OVM_StateManagerFactory")).create(address(this));
+        ovmBondManager = OVM_BondManager(resolve("OVM_BondManager"));
```
kfichter, line 83: Hmmm now that I'm thinking about it, do we also want to include the users who call `initializeFraudVerification` and `finalizeFraudVerification` in the reward? @ben-chain @karlfloersch

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,201 @@
...
+    string constant TOO_EARLY = "BondManager: Too early to finalize your withdrawal";
+
+    string constant ONLY_OWNER = "BondManager: Only the contract's owner can call this function";
+    string constant ONLY_TRANSITIONER = "BondManager: Only the transitioner for this pre-state root may call this function";
+    string constant ONLY_FRAUD_VERIFIER = "BondManager: Only the fraud verifier may call this function";
+    string constant ONLY_STATE_COMMITMENT_CHAIN = "BondManager: Only the state commitment chain may call this function";
+}
+
+contract OVM_BondManager is Lib_AddressResolver {
+    /// Owner used to bump the security bond size
```
kfichter, line 31: Small request to reformat to use the same format as other contracts (i.e.,:

```
/***********
 * Structs *
 ***********/

// whatever

/*************
 * Constants *
 *************/

// whatever

```

etc. 

Primarily just to keep the contract structure stable for the sake of auditors.

kfichter (commented):
Looking good!! Added some stylistic feedback but I think this is functionally correct.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
gakonst: https://github.com/ethereum-optimism/contracts-v2/pull/3/commits/932582a2f41baee64ed338bb1e3c2c4ac443bcfb done, check the commit message for adjusting your deploy scripts

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
gakonst: As discussed, I'd like to avoid this optimization: Doing the changes as suggested seems to break my current tests.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
karlfloersch: This seems reasonable!

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
kfichter: Ok we're just going to go with your current solution, marking this as resolved.

ben-chain (commented):
Lookin' good, woot woot!!!

One wishlist item for the current structure is that the process does not take into account the amount of gas actually spent by the folks proving fraud.  I could imagine a 

```
modifier contributesToFraud() {
    uint startGas = gasleft();
    _;
   uint gasSpent = startGas - gasleft();
   bondManager.recordGasSpent(msg.sender, gasSpent, ...)
}
```

Which would potentially take care of this cleanly.  @gakonst Thoughts on this? I don't think it's strictly necessary to get this merged in.

contracts/optimistic-ethereum/OVM/OVM_BondManager.sol
```diff
@@ -0,0 +1,223 @@
...
+    function setRequiredCollateral(uint256 newValue) public {
+        require(newValue > requiredCollateral, Errors.LOW_VALUE);
+        require(msg.sender == owner, Errors.ONLY_OWNER);
+        requiredCollateral = newValue;
+    }
+
+    /// Checks if the user is collateralized for the batchIndex
+    function isCollateralized(address who) public view returns (bool) {
+        require(bonds[who].locked >= requiredCollateral, Errors.NOT_ENOUGH_COLLATERAL);
+        return true;
```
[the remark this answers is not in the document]
gakonst: eh I prefer it this way, similarly to how most ERC20's transfers return true but revert internally, instead of returning false on error conditions. It just removes the burden from the consumer calling this function

contracts/optimistic-ethereum/OVM/OVM_FraudContributor.sol
```diff
@@ -0,0 +1,19 @@
...
+import { OVM_BondManager } from "./OVM_BondManager.sol";
+
+/// Minimal contract to be inherited by contracts consumed by users that provide
+/// data for fraud proofs
+contract OVM_FraudContributor {
+    OVM_BondManager internal ovmBondManager;
+
+    /// Decorate your functions with this modifier to store how much total gas was
+    /// consumed by the sender, to reward users fairly
+    modifier contributesToFraud(bytes32 preStateRoot) {
```
kfichter, line 13: Maybe worth renaming to `contributesToFraudProof` instead 😂

kfichter (approved):
LGTM

PR #5, opened 2020-10-12
Add timestamp and auth to message passing contracts

kfichter:
Adds `timestamp` to `ChainBatchHeader` and uses the value to assert that a message is outside of the fraud proof window. Also adds authentication to the message passing contracts via `AddressManager`/`AddressResolver` pattern.

### DIFF
--- a/contracts/optimistic-ethereum/mockOVM/bridge/mockOVM_CrossDomainMessenger.sol
+++ b/contracts/optimistic-ethereum/mockOVM/bridge/mockOVM_CrossDomainMessenger.sol
@@ -29,6 +29,7 @@ contract mockOVM_CrossDomainMessenger is OVM_BaseCrossDomainMessenger {
      **********************/
 
     ReceivedMessage[] internal fullReceivedMessages;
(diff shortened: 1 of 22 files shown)


ben-chain (approved):
👍👍👍

contracts/optimistic-ethereum/mockOVM/bridge/mockOVM_CrossDomainMessenger.sol
```diff
@@ -29,6 +29,7 @@ contract mockOVM_CrossDomainMessenger is OVM_BaseCrossDomainMessenger {
      **********************/
 
     ReceivedMessage[] internal fullReceivedMessages;
+    address public targetMessengerAddress;
```
K-Ho, line 32: Let's make this internal, since it's no longer in the real CrossDomainMessengers

K-Ho (approved):
Lgtm! Somewhat tangential, but do we need to add functionality to the contract deployment script to deploy the L2CrossDomainMessenger and an AddressResolver to L2?

PR #6, opened 2020-10-12
Made StateManager usage authenticated

kfichter:
Makes `OVM_StateManager` usage within `run` authenticated.


ben-chain (approved):
Woot woot, LGTM!

PR #7, opened 2020-10-13
Add Safety Checker

K-Ho:
Add the Safety Checker, adding the updated KALL and KOPY checks, blacklisting REVERT.
Also makes sure the SafetyChecker is smocked in the other tests.


ben-chain (approved):
LGTM!

PR #8, opened 2020-10-13
Restructure batch submission & add ring buffer

karlfloersch:
This PR overhauls the batch submission logic & adds a ring buffer.

### Change to batch submission
Now the CTC submits batches which consist of some number of transactions, each one with a timestamp / block number. Additionally, the ability to submit a single batch with many queue transactions is now possible. Or sequencer batches with queue and sequencer transactions interspersed.


contracts/optimistic-ethereum/libraries/utils/Lib_TimeboundRingBuffer.sol
```diff
@@ -0,0 +1,188 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.7.0;
+
+/* Logging */
+import { console } from "@nomiclabs/buidler/console.sol";
```
tynes, line 5: This should be removed before this is merged right?

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.spec.ts
```diff
@@ -9,13 +9,19 @@ import { smockit, MockContract } from '@eth-optimism/smock'
 import {
   makeAddressManager,
   setProxyTarget,
-  getEthTime,
-  setEthTime,
-  NON_NULL_BYTES32,
   FORCE_INCLUSION_PERIOD_SECONDS,
-  ZERO_ADDRESS,
+  increaseEthTime,
+  // NON_NULL_BYTES32,
```
tynes, line 14: Dead code

tynes:
The L1ToL2 event still needs to be implemented, not sure if out of scope of this PR

contracts/optimistic-ethereum/libraries/utils/Lib_TimeboundRingBuffer.sol
karlfloersch: Oh yes it should be! Good catch

karlfloersch:
Merged with https://github.com/ethereum-optimism/contracts-v2/pull/9

PR #17, opened 2020-10-15
Emit tx index

karlfloersch:
Add the total elements to the event emission to make it easier to sync.


kfichter (approved):
wow very approved

PR #21, opened 2020-10-16
Fix JUMPI inputs ordering

ben-chain:
Fixes a small bug in the safety checked logic where the input to `JUMPI` switched the `condition` and `destination`.


K-Ho (approved):
LGTM!

PR #27, opened 2020-10-20
Require height specification in SCC

kfichter:
## Description
PR changes `StateCommitmentChain` to require that users provide a height for their batches, also introduces a window during which only the `sequencer` can submit state batches as a temporary way to make breaking bugs less annoying.

## Contributing Agreement


contracts/optimistic-ethereum/OVM/chain/OVM_StateCommitmentChain.sol
```diff
@@ -14,6 +14,8 @@ import { iOVM_StateCommitmentChain } from "../../iOVM/chain/iOVM_StateCommitment
 import { iOVM_CanonicalTransactionChain } from "../../iOVM/chain/iOVM_CanonicalTransactionChain.sol";
 import { iOVM_BondManager } from "../../iOVM/verification/iOVM_BondManager.sol";
 
+/* Logging Imports */
+import { console } from "@nomiclabs/buidler/console.sol";
```
tynes, line 18: This should be removed before merge

PR #34, opened 2020-10-25
Allow easy emptying of the queue

karlfloersch:
## Description
Fixes a slight race condition where if you attempt to empty the queue with a number of elements that is larger than the total number of elements in the queue, you will have your tx revert. Because of this we had a mild race condition where if two people append queue batches, one might fail.

Additionally there was some weird behavior where you couldn't verify an element unless it was passed the force inclusion period. This isn't fatal but increases client complexity.

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -242,9 +242,23 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
             "Must append more than zero transactions."
         );
 
-        uint40 nextQueueIndex = getNextQueueIndex();
         bytes32[] memory leaves = new bytes32[](_numQueuedTransactions);
+        uint40 nextQueueIndex = getNextQueueIndex();
+        uint40 queueLength = _getQueueLength();
+
         for (uint256 i = 0; i < _numQueuedTransactions; i++) {
+            if (nextQueueIndex >= queueLength
+                || (
+                    msg.sender != sequencer
+                    && getQueueElement(nextQueueIndex).timestamp + forceInclusionPeriodSeconds > block.timestamp
+                )
+            ) {
+                // Quit early because nextQueueIndex is larger than the queue or queue
+                // transactions is not allowed to be submitted during the sequencer inclusion period.
+                require(i > 0, "No appendable queue elements.");
+                _numQueuedTransactions = i;
+                break;
+            }
             leaves[i] = _getQueueLeafHash(nextQueueIndex);
             nextQueueIndex++;
         }
@@ -313,6 +327,7 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         uint32 transactionIndex = 0;
         uint32 numSequencerTransactionsProcessed = 0;
         uint40 nextQueueIndex = getNextQueueIndex();
+        uint40 queueLength = _getQueueLength();
 
         for (uint32 i = 0; i < numContexts; i++) {
             BatchContext memory context = _getBatchContext(i);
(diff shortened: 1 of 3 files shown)


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -521,6 +529,21 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         );
     }
 
+    /**
+     * Retrieves the length of the queue.
+     * @return Length of the queue.
+     */
+    function _getQueueLength(
+    )
```
tynes, line 537: Unnecessary newline

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -521,6 +529,21 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
+        internal
+        view
+        returns (
+            uint40
+        )
+    {
+        return queue.getLength()/2;
```
karlfloersch, line 544: We have to divide by 2 because we store 2 values for every queue element

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -237,14 +237,26 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
 
-        uint40 nextQueueIndex = getNextQueueIndex();
         bytes32[] memory leaves = new bytes32[](_numQueuedTransactions);
+
         for (uint256 i = 0; i < _numQueuedTransactions; i++) {
+            if (msg.sender != sequencer) {
+                Lib_OVMCodec.QueueElement memory el = getQueueElement(nextQueueIndex);
+                require(
+                    el.timestamp + forceInclusionPeriodSeconds < block.timestamp,
+                    "Queue transactions cannot be submitted during the sequencer inclusion period."
```
ben-chain, line 257: maybe `"Queue transactions can only be submitted by sequencer during force inclusion period"`?

ben-chain: Can also just be 
```
require(
                   msg.sender == sequencer ||
                    getQueueElement(nextQueueIndex).timestamp + forceInclusionPeriodSeconds < block.timestamp,
                   "Queue transactions can only be submitted by sequencer during force inclusion period."
```
but up to you on style there.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -237,14 +237,26 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         override
         public
     {
+        uint40 nextQueueIndex = getNextQueueIndex();
+        uint40 queueLength = _getQueueLength();
+        if (queueLength - nextQueueIndex < _numQueuedTransactions) {
+            _numQueuedTransactions = queueLength - nextQueueIndex;
+        }
```
ben-chain, line 244: Would be a lot prettier to have `Math.min` if there's an easy URL import

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
tynes: You should add this comment to the source code

ben-chain (approved):
Left a couple small style points but LGTM!

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
karlfloersch: Re. the revert message -- I inherited the revert message from the previous implementation so I'm tempted to not change it so I don't change the tests

tynes (approved):

PR #35, opened 2020-10-26
deployment: script

tynes:
## Description

Adds a deployment script for the contracts to be used in test automation

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement


bin/deploy.js
```diff
@@ -0,0 +1,67 @@
...
+const contracts = require('../src/index.ts');
+const { providers, Wallet } = require('ethers')
+const { JsonRpcProvider } = providers;
+
+const key = process.env.DEPLOYER_PRIVATE_KEY;
+const sequencerKey = process.env.SEQUENCER_PRIVATE_KEY;
+const web3Url = process.env.L1_NODE_WEB3_URL || 'http://127.0.0.1:8545';
+const MIN_TRANSACTION_GAS_LIMIT = process.env.MIN_TRANSACTION_GAS_LIMIT || 0;
+const MAX_TRANSACTION_GAS_LIMIT = process.env.MAX_TRANSACTION_GAS_LIMIT || 1000000000;
+const MAX_GAS_PER_QUEUE_PER_EPOCH = process.env.MAX_GAS_PER_QUEUE_PER_EPOCH || 1000000000000;
```
karlfloersch, line 12: Curious if @ben-chain has input here but my guess is that this default is a bit too high when coupled with the `600` second epoch time. I think a safer one is `500000000` , but maybe even safer is `250000000`

bin/deploy.js
```diff
@@ -0,0 +1,67 @@
...
+const SECONDS_PER_EPOCH = process.env.SECONDS_PER_EPOCH = 600;
+let WHITELIST_OWNER = process.env.WHITELIST_OWNER;
+const WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT = process.env.WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT || false;
```
karlfloersch, line 15: I think this should default to `true` because most of the time we want chains to allow arbitrary contracts

bin/deploy.js
```diff
@@ -0,0 +1,67 @@
...
+const FORCE_INCLUSION_PERIOD_SECONDS = process.env.FORCE_INCLUSION_PERIOD_SECONDS || 1 << 28;
```
karlfloersch, line 16: I think this might be fine as a default for testing, but we'd probably want something closer to 15 minutes or 30 minutes in practice... or fewer minutes even but 30 mins is probably fine at first & we can reduce it over time down to our eventual 2 minute target once we've tested more throughly

karlfloersch (approved):
LGTM!

bin/deploy.js
karlfloersch: > 250000000

This is the one to use

PR #38, opened 2020-10-28
Add getPendingQueueElements

karlfloersch:
## Description
Add getPendingQueueElements

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -170,6 +170,20 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
     /**
      * @inheritdoc iOVM_CanonicalTransactionChain
      */
+    function getNumPendingQueueElements()
+        override
+        public
+        view
+        returns (
+            uint40
+        )
+    {
+        return  _getQueueLength() - getNextQueueIndex();
+    }
+
+    /**
+     * @inheritdoc iOVM_CanonicalTransactionChain
+     */
     function enqueue(
         address _target,
         uint256 _gasLimit,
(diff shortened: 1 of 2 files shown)


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -248,15 +262,14 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         override
         public
     {
-        uint40 nextQueueIndex = getNextQueueIndex();
-        uint40 queueLength = _getQueueLength();
-        _numQueuedTransactions = Math.min(_numQueuedTransactions, queueLength - nextQueueIndex);
+        _numQueuedTransactions = Math.min(_numQueuedTransactions, getNumPendingQueueElements());
```
ben-chain, line 265: Wouldn't this be a sign that there is a reorg in the honest sequencer case?  Could even throw if _numqueued exceeds numpending.

ben-chain (approved):
left a comment/question but LGTM!

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
karlfloersch: Ah you’re right but this is used a little differently. So when the sequencer goes down and we need to clear the queue, the batch submitter just blindly calls this with a massive number to make sure that the queue is for sure emptied.

karlfloersch: And then this is also useful for normal nodes if they ever want to clear the queue without any race conditions

ben-chain: great, ship it then!

PR #53, opened 2020-11-09
Add messenger proxy

karlfloersch:
## Description
Adds a proxy for the L1Messenger!

## Questions
- Any objections with the way that I tested the proxy? Basically just used it for the L1Messenger tests.

### Fixes
- Fixes https://github.com/ethereum-optimism/roadmap/issues/147

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
@@ -42,8 +42,6 @@ contract OVM_L1CrossDomainMessenger is iOVM_L1CrossDomainMessenger, OVM_BaseCros
     {
         require(address(libAddressManager) == address(0), "L1CrossDomainMessenger already intialized.");
         libAddressManager = Lib_AddressManager(_libAddressManager);
-        ovmCanonicalTransactionChain = iOVM_CanonicalTransactionChain(resolve("OVM_CanonicalTransactionChain"));
-        ovmStateCommitmentChain = iOVM_StateCommitmentChain(resolve("OVM_StateCommitmentChain"));
     }
 
 
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
@@ -5,6 +5,7 @@ pragma experimental ABIEncoderV2;
 /* Library Imports */
 import { Lib_OVMCodec } from "../../libraries/codec/Lib_OVMCodec.sol";
 import { Lib_AddressResolver } from "../../libraries/resolver/Lib_AddressResolver.sol";
+import { Lib_AddressManager } from "../../libraries/resolver/Lib_AddressManager.sol";
 import { Lib_SecureMerkleTrie } from "../../libraries/trie/Lib_SecureMerkleTrie.sol";
 
 /* Interface Imports */
@@ -40,6 +41,7 @@ contract OVM_L1CrossDomainMessenger is iOVM_L1CrossDomainMessenger, OVM_BaseCros
         public
     {
         require(address(libAddressManager) == address(0), "L1CrossDomainMessenger already intialized.");
+        libAddressManager = Lib_AddressManager(_libAddressManager);
         ovmCanonicalTransactionChain = iOVM_CanonicalTransactionChain(resolve("OVM_CanonicalTransactionChain"));
         ovmStateCommitmentChain = iOVM_StateCommitmentChain(resolve("OVM_StateCommitmentChain"));
     }
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
@@ -25,11 +25,24 @@ contract OVM_L1CrossDomainMessenger is iOVM_L1CrossDomainMessenger, OVM_BaseCros
      ***************/
 
     /**
+     * Pass a default zero address to the address resolver. This will be updated when initialized.
+     */
+    constructor()
+        Lib_AddressResolver(address(0))
+    {}
+
+    /**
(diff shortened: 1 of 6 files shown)


test/contracts/OVM/bridge/OVM_L1CrossDomainMessenger.spec.ts
```diff
@@ -89,9 +103,10 @@ describe('OVM_L1CrossDomainMessenger', () => {
 
   let OVM_L1CrossDomainMessenger: Contract
   beforeEach(async () => {
-    OVM_L1CrossDomainMessenger = await Factory__OVM_L1CrossDomainMessenger.deploy(
-      AddressManager.address
-    )
+    const xDomainMessenerImpl = await Factory__OVM_L1CrossDomainMessenger.deploy()
+    // We use an upgradable proxy for the XDomainMessenger--deploy & set up the proxy.
+    OVM_L1CrossDomainMessenger = await deployProxyXDomainMessenger(AddressManager, xDomainMessenerImpl)
+    await OVM_L1CrossDomainMessenger.initialize(AddressManager.address)
```
karlfloersch, line 109: I'm using this test suite to sanity check the proxy implementation -- ideally there would be unit tests for the proxy

test/contracts/libraries/trie/Lib_SecureMerkleTrie.spec.ts
```diff
@@ -7,7 +7,7 @@ import { Contract } from 'ethers'
 /* Internal Imports */
 import { TrieTestGenerator } from '../../../helpers'
 
-const NODE_COUNTS = [1, 2, 128, 256, 512, 1024, 2048, 4096]
+const NODE_COUNTS = [1, 2, 128]
```
karlfloersch, line 10: Just set this to speed up our test suite

contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
```diff
@@ -24,12 +25,24 @@ contract OVM_L1CrossDomainMessenger is iOVM_L1CrossDomainMessenger, OVM_BaseCros
...
+     */
+    constructor()
+        Lib_AddressResolver(address(0))
+    {}
+
     /**
      * @param _libAddressManager Address of the Address Manager.
      */
-    constructor(
+    function initialize(
```
K-Ho, line 38: Why do we need to initialize this address instead of in the constructor?

contracts/optimistic-ethereum/libraries/resolver/Lib_ResolvedDelegateProxy.sol
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @title Lib_ResolvedDelegateProxy
+ */
+contract Lib_ResolvedDelegateProxy {
+
+    /*************
+     * Variables *
+     *************/
+
+    mapping(address=>string) private implementationName;
```
K-Ho, line 16: Add comment for this wacky storage slot key hacking

contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
K-Ho: nvm! I get it - it's to ensure that all storage slots are set in the proxy

K-Ho (approved):
lgtm!

test/contracts/OVM/bridge/OVM_L1CrossDomainMessenger.spec.ts
```diff
@@ -34,6 +34,20 @@ const getXDomainCalldata = (
...
+  l1XDomainMessenger: Contract
+): Promise<Contract> => {
+  await addressManager.setAddress(
+    'OVM_L1CrossDomainMessenger',
+    l1XDomainMessenger.address
+  )
+  const proxy = await (await ethers.getContractFactory(
+      'Lib_ResolvedDelegateProxy'
+  )).deploy(addressManager.address, 'OVM_L1CrossDomainMessenger')
+  return l1XDomainMessenger.attach(proxy.address)
```
K-Ho, line 48: Just double checking that this will just be a contract using the L1CrossDomainMessenger interface, but the address of the proxy?

karlfloersch: Yep exactly! I didn't do this at first & instead constructed the bytecode to send manually, but then realized that hilariously because the proxy just forwards everything along, I can just highjack the normal Ethers contract objects to make calls to the proxy.

src/contract-deployment/config.ts
```diff
@@ -43,7 +43,17 @@ export const makeContractDeployConfig = async (
...
-      params: [AddressManager.address],
+      params: [],
+    },
+    Proxy__OVM_L1CrossDomainMessenger: {
+      factory: getContractFactory('Lib_ResolvedDelegateProxy'),
+      params: [AddressManager.address, 'OVM_L1CrossDomainMessenger'],
+      afterDeploy: async (contracts): Promise<void> => {
+        const xDomainMessenger = getContractFactory(
+          'OVM_L1CrossDomainMessenger'
+        ).attach(contracts.Proxy__Proxy__OVM_L1CrossDomainMessenger.address)
```
tynes, line 54: Why is it `Proxy__Proxy`?

test/contracts/libraries/trie/Lib_SecureMerkleTrie.spec.ts
tynes: Maybe too aggressive in removing test coverage? I'm not super familiar with the implementation so it would be best to get an ok from the people who wrote the code

src/contract-deployment/config.ts
karlfloersch: 😭 Lemme add an issue for removing the `Proxy__` prefix. Right now all contracts are incorrectly prefixed with `Proxy__` so we need to remove that prefix, but think that should be in another PR

karlfloersch: https://github.com/ethereum-optimism/contracts-v2/issues/55 -- here we go!

PR #64, opened 2020-11-16
Bond Manager Deposit bugfix

gakonst:
- Deposits `requiredCollateral` every time
- Removes the `setRequiredCollateral` admin function. This removes the `owner` role from the contract (good!) and it also prevents inconsistencies which would require that we track more variables such as the ones described in https://github.com/ethereum-optimism/contracts-v2/commit/e57df4c638da8082741d070d5a355cabff0bab97 's commit message
- Removes the redundant `batchIndex` parameter

@kfichter @karlfloersch @ben-chain


ben-chain (approved):
Appreciate simplicity, LGTM!

PR #65, opened 2020-11-16
Reward earliest fraud proof only

### LINKED ISSUE #43: Pay bond to earliest fraud proof
**Describe the bug**
The attack is as follows:
1. Attacker commits fraud at time T and T+10.
2. Attacker proves fraud at time T+10, getting 1/2 of their bond back.
3. Verifier now needs to prove fraud at time T to protect their funds, but they do not get a reward for doing so.

**Steps to reproduce**
http://www.gambit-project.org/
lulz jk

**Expected behavior**
The bond should be paid out to the people who contributed to the **earliest** fraud proof.

**Solution**
Once fraud is proven at `T+10`, start withdrawal period where `earliestFraud=T+10`  . Then once fraud is proven a second time, `earliestFraud = min(currentEarliestFraud, T)` . Of course The new fraud proof at time `T` is now the `earliestFraud`.
At the end of the week, if you contributed to the `earliestFraud` fraud proof, you get the payout!

Shout out @kfichter for figuring this out.

gakonst:
- Base: https://github.com/ethereum-optimism/contracts-v2/pull/64
- Fixes #43 

Instead of claiming per state root, users now claim per data publisher. Rewards are distributed ONLY from the earliest proposed publisher state. We allow claiming only after T, giving time for earlier frauds to finalize and be rewarded

@kfichter @karlfloersch @ben-chain


ben-chain (approved):
Awesome, LGTM!!

PR #67, opened 2020-11-17
Upgrade/dual compile

axchu:
## Description
compiles ovm, evm, or both for different contracts as specified.
re-architects the artifacts & cache folders into ovm-artifacts, evm-artifacts, ovm-cache, evm-cache.

## Questions
there are some tests failing atm (scripts that depend on the old artifacts folder)
deployment is not updated yet

## Metadata
is upgraded to hardhat (from upgrade/hardhat) 
needs smock upgraded to hardhat as well (from upgrade/hardhat in smock repo)


kfichter:
Going to close this since we have a temporary way of doing the dual compile already merged into master. We can work on a more long-term solution after minnet.

PR #71, opened 2020-11-19
[WIP] State Manager Gas Consumption Script

ben-chain:
## Description
This PR adds a little script which meters and prints the gas consumption of the `OVM_StateManager` methods under various conditions.
## Questions
Any cases I missed?

### Fixes
- Fixes roadmap # 22

## Contributing Agreement


contracts/test-helpers/Helper_GasMeasurer.sol
```diff
@@ -0,0 +1,32 @@
...
+        assembly {
+            calldataStart := add(_data,0x20)
+            calldataLength := mload(_data)
+        }
+
+        bool success;
+        assembly {
+            gasBefore := gas()
+            success := call(gas(), _target, 0, calldataStart, calldataLength, 0, 0)
+            gasAfter := gas()
```
ben-chain, line 26: I'm not sure exactly how this overhead works, but presumably will need to subtract at minimum the cost of `GAS + `CALL`, and possibly more, to get this to be the value we consume in geth.

kfichter:
First pass looks good to me. Will do a more thorough review of all the test cases tomorrow just to double check that the paths are covered. 👍 

I'll think on the figuring out how to get the exact gas cost. Maybe we can just debug and look at the opcodes between the gas recording points.

test/contracts/OVM/execution/OVM_StateManager.gas-spec.ts
```diff
@@ -0,0 +1,527 @@
...
+                NON_NULL_BYTES32
+            ],
+            async () => {
+                await OVM_StateManager.initPendingAccount(NON_ZERO_ADDRESS)
+            }
+        )
+    })
+
+    describe('getContractStorage', () => {
+        // confirm with kelvin that this covers all cases
```
kfichter, line 408: Yep LGTM

kfichter (approved):
Looks good. I think this covers everything.

PR #78, opened 2020-11-24
Fix ovmSTATICCALL behavior

ben-chain:
## Description
This PR updates the behavior of `ovmSTATICCALL` so that it more closely aligns with the EVM's behavior.  We had made it so that static violations make the original staticcall at whatever level higher up fail, but in reality, violation of the static context in the EVM only makes that particular message throw an exception.

## Questions
N/A 

### Fixes
- Fixes roadmap # 188, OZ H04

## Contributing Agreement


kfichter:
Oh interesting. I didn't realize this was the EVM behavior of STATICCALL. Cross referencing with EthereumJS, just a second.

kfichter:
Yep, this PR has the correct behavior. Approving.

kfichter (approved):
LGTM

PR #79, opened 2020-11-24
Audit/oz co2 match transition indexing 187

axchu:
## Questions
- 
-
-

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/verification/OVM_FraudVerifier.sol
+++ b/contracts/optimistic-ethereum/OVM/verification/OVM_FraudVerifier.sol
@@ -26,7 +26,7 @@ contract OVM_FraudVerifier is Lib_AddressResolver, OVM_FraudContributor, iOVM_Fr
      *******************************************/
 
     mapping (bytes32 => iOVM_StateTransitioner) internal transitioners;
-    
+
 
     /***************
      * Constructor *
@@ -95,7 +95,7 @@ contract OVM_FraudVerifier is Lib_AddressResolver, OVM_FraudContributor, iOVM_Fr
         if (_hasStateTransitioner(_preStateRoot)) {
             return;
         }
-        
+
         iOVM_StateCommitmentChain ovmStateCommitmentChain = iOVM_StateCommitmentChain(resolve("OVM_StateCommitmentChain"));
         iOVM_CanonicalTransactionChain ovmCanonicalTransactionChain = iOVM_CanonicalTransactionChain(resolve("OVM_CanonicalTransactionChain"));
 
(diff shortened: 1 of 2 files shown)


test/contracts/OVM/verification/OVM_FraudVerifier.spec.ts
```diff
@@ -178,6 +178,54 @@ describe('OVM_FraudVerifier', () => {
...
+              preStateDummyBatchProof,
+              DUMMY_OVM_TRANSACTIONS[0],
+              DUMMY_TX_CHAIN_ELEMENTS[0],
+              DUMMY_BATCH_HEADERS[0],
+              transactionDummyBatchProof
+            )
+          ).to.be.revertedWith('Invalid transaction inclusion proof.')
+        })
+      })
+    })
```
ben-chain, line 207: I think that instead of this test, what we're looking for are tests that get `revertedWith("Pre-state root global index must equal to the transaction root global index")`, since we've added that assertion in the contract. So this would be the case where the inclusion proofs for the state root and the transaction both pass, but they are for non-matching indices.

contracts/optimistic-ethereum/OVM/verification/OVM_FraudVerifier.sol
```diff
@@ -118,6 +118,11 @@ contract OVM_FraudVerifier is Lib_AddressResolver, OVM_FraudContributor, iOVM_Fr
             "Invalid transaction inclusion proof."
         );
 
+        require (
+            _preStateRootBatchHeader.prevTotalElements + _preStateRootProof.index == _transactionBatchHeader.prevTotalElements + _transactionProof.index,
+            "Pre-state root global index must equal to the transaction root global index"
```
ben-chain, line 123:
```suggestion
            "Pre-state root global index must equal to the transaction root global index."
```

test/contracts/OVM/verification/OVM_FraudVerifier.spec.ts
```diff
@@ -178,6 +178,54 @@ describe('OVM_FraudVerifier', () => {
...
+
+    describe('when provided an invalid transaction', () => {
```
ben-chain, line 209: I think these is a good test to add--let's get a little more specific though!  I think if we did two cases:
- `describe('when provided an invalid transaction proof')`
- `describe('when provided an invalid state root proof')`
and had `revertedWith` checks on each, that would be super dope.

ben-chain (approved):
Woot woot!  Code looks right to me, left some small nits and request to improve tests--approving in advance of those changes coming in as contract looks right.

test/contracts/OVM/verification/OVM_FraudVerifier.spec.ts
axchu: so I didn't see it before, but I think this is already done...
line 105 is 'invalid prestate root proof' and 'valid tx proof'
line 134 is 'valid prestate root roof' and 'invalid tx proof'

PR #83, opened 2020-11-25
chainId check

axchu:
### Fixes
https://github.com/ethereum-optimism/roadmap/issues/191

## Contributing Agreement

### DIFF
--- a/test/helpers/codec/encoding.ts
+++ b/test/helpers/codec/encoding.ts
@@ -121,7 +121,7 @@ export const signNativeTransaction = async (
   let [v, r, s] = getSignedComponents(transactionSignature).map((component) => {
     return remove0x(component)
   })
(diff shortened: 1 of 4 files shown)


K-Ho (approved):
LGTM!

test/helpers/codec/encoding.ts
```diff
@@ -121,7 +121,7 @@ export const signNativeTransaction = async (
   let [v, r, s] = getSignedComponents(transactionSignature).map((component) => {
     return remove0x(component)
   })
-  v = '0' + (parseInt(v, 16) - 420 * 2 - 8 - 27)
+  v = '0' + (parseInt(v, 16) - transaction.chainId * 2 - 8 - 27)
```
tynes, line 124: I don't think that this will work for all possible chainIds, if the right hand side of the string concat results in an even length string, then v will be odd length

PR #84, opened 2020-11-25
Fix RLPWriter Memory Allocation

ben-chain:
## Description
Adds a `_malloc` function for clearing memory at the free pointer, and adds a test which would fail for `writeAddress` without it.

## Questions
Which functions need this applied? All of them?

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/libraries/rlp/Lib_RLPWriter.sol
+++ b/contracts/optimistic-ethereum/libraries/rlp/Lib_RLPWriter.sol
@@ -89,7 +89,7 @@ library Lib_RLPWriter {
             bytes memory _out
         )
     {
-        _malloc(1);
+        _malloc(0x20);
         bytes memory inputBytes;
         assembly {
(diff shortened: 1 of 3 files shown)


contracts/optimistic-ethereum/libraries/rlp/Lib_RLPWriter.sol
```diff
@@ -308,4 +309,19 @@ library Lib_RLPWriter {
 
         return flattened;
     }
+
+    function _malloc(
+        uint _numWords
+    )
+        internal
```
kfichter, line 316: Should be marked `private` for consistency but otherwise LGTM.

kfichter (commented):
Looks good. Can approve if you need it, could use a few more tests if it's not too difficult to demonstrate the effect.

ben-chain:
@kfichter tests on this are gonna come in downstream via existing audit tests -- can you approve?

PR #85, opened 2020-12-01
Audit/dapphub b02 gas relaying check

axchu:
### Fixes
- Fixes https://github.com/ethereum-optimism/roadmap/issues/221
- Fixes https://github.com/ethereum-optimism/roadmap/issues/222

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
+++ b/contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
@@ -18,6 +18,7 @@ contract OVM_ECDSAContractAccount is iOVM_ECDSAContractAccount {
 
     using SafeMath for uint256;
     address constant ETH_ERC20_ADDRESS = 0x4200000000000000000000000000000000000006;
+    uint256 constant EXECUTION_VALIDATION_GAS_OVERHEAD = 25000; // TODO: should be the amount sufficient to cover the gas costs of all of the transactions up to and including the CALL/CREATE which forms the entrypoint of the transaction.
 
     /********************
      * Public Functions *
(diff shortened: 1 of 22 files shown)


axchu:
EXECUTION_VALIDATION_GAS_OVERHEAD is a band-aid, can/should be improved upon.  
It should be the amount sufficient to cover the gas costs of all of the transactions up to and including the CALL/CREATE which forms the entrypoint of the transaction.

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
```diff
@@ -76,6 +79,12 @@ contract OVM_ECDSAContractAccount is iOVM_ECDSAContractAccount {
             "Transaction nonce does not match the expected nonce."
         );
 
+        // Need to make sure that the gas is sufficient to execute the transaction.
+        Lib_SafeExecutionManagerWrapper.safeREQUIRE(
+           gasleft() >= SafeMath.add(decodedTx.gasLimit, EXECUTION_VALIDATION_GAS_OVERHEAD),
```
xwvvvvwx, line 84: Is the openzeppelin safeMath ok to use here? I think that will call `require` directly instead of `safeREQUIRE`?

axchu: yes you are right...

axchu: solved instead with Lib_SafeMathWrapper

contracts/optimistic-ethereum/libraries/wrappers/Lib_SafeMathWrapper.sol
```diff
@@ -0,0 +1,166 @@
+// SPDX-License-Identifier: MIT
```
ben-chain, line 1: Can we add some credit to where we pulled from here?

test/contracts/OVM/accounts/OVM_ECDSAContractAccount.spec.ts
```diff
@@ -227,5 +229,33 @@ describe('OVM_ECDSAContractAccount', () => {
...
+
+    it(`should revert on insufficient gas`, async () => {
+      const alteredInsufficientGasTx = {
+        ...DEFAULT_EIP155_TX,
+        gasLimit : 88888888
+      }
+      const message = serializeNativeTransaction(alteredInsufficientGasTx)
+      const sig = await signNativeTransaction(wallet, alteredInsufficientGasTx)
+
+      await callPrecompile(
```
ben-chain, line 241: Might be worth adding a custom `gasLimit` field to this test function so that we can explicitly assert the gas failure, not just by setting an implicitly too high of `gasLimit`.

ben-chain (approved):
LGTM, left a couple small requests but I think the logic is solid 👍

PR #94, opened 2020-12-03
Fix run gas Lower Bound

ben-chain:
## Description
This PR ensures that the `run` function is supplied with sufficient gas to execute in full, including accounting for EIP150.

### Fixes
- Fixes roadmap#248 

## Contributing Agreement


contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -134,20 +134,25 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-        // Check whether we need to start a new epoch, do so if necessary.
-        _checkNeedsNewEpoch(_transaction.timestamp);
-
         // Make sure the transaction's gas limit is valid. We don't revert here because we reserve
         // reverts for INVALID_STATE_ACCESS.
         if (_isValidGasLimit(_transaction.gasLimit, _transaction.l1QueueOrigin) == false) {
             return;
         }
 
+        // We require gas to complete the logic here in run() before/after execution,
```
ben-chain, line 143: @kfichter let's make sure we nail down the order of these checks when the time comes, I think you also have a diff on the `initializeContext`.

ben-chain:
Just realized something -- I think I should add a test on this bad boy with a transaction whose `data` field is the `MAX_OVM_TX_SIZE` so that we know the pad is sufficient to cover the worst-case gas consumption during `run`.

ben-chain:
@kfichter need another review on this since merging master--thank you!

PR #97, opened 2020-12-07
Various fixes found while working on the fraud proving service

kfichter:
## Description
Helloooo. This PR contains a number of different fixes that I found while working on the fraud proving service. Marking as draft until I have tests passing, but could still use review on the changes to the contracts.

## Contributing Agreement


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -791,8 +791,8 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         require(
             _transaction.blockNumber        == _txChainElement.blockNumber
             && _transaction.timestamp       == _txChainElement.timestamp
-            && _transaction.entrypoint      == resolve("OVM_DecompressionPrecompileAddress")
-            && _transaction.gasLimit        == gasLimit
+            && _transaction.entrypoint      == 0x4200000000000000000000000000000000000005
+            && _transaction.gasLimit        == 7000000
```
kfichter, line 795: This was temporary, need to revert the change.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -134,6 +134,9 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
             "Only authenticated addresses in ovmStateManager can call this function"
         );
 
+        // Initialize the execution context.
+        _initContext(_transaction);
+
```
kfichter, line 139: Context has to be initialized *before* we call `_checkNeedsNewEpoch` or the whole transaction fails.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -477,6 +477,8 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
             address(proxyEOA),
             keccak256(Lib_EthUtils.getCode(address(proxyEOA)))
         );
+
+        _setAccountNonce(eoa, 0);
```
kfichter, line 481: EOA are expected to have a nonce of 0, but `_commitPendingAccount` sets the nonce to 1. So we reset it here.

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -315,7 +315,7 @@ contract OVM_StateManager is iOVM_StateManager {
         )
     {
         return _testAndSetItemState(
-            keccak256(abi.encodePacked(_address)),
+            _getAccountItemHash(_address),
```
kfichter, line 318: Wanted to simplify this logic.

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -392,7 +392,35 @@ contract OVM_StateManager is iOVM_StateManager {
     {
         return totalUncommittedAccounts;
     }
-    
+
+    function wasAccountChanged(
```
kfichter, line 396: Useful functions for the client.

contracts/optimistic-ethereum/OVM/verification/OVM_FraudVerifier.sol
```diff
@@ -122,10 +122,10 @@ contract OVM_FraudVerifier is Lib_AddressResolver, OVM_FraudContributor, iOVM_Fr
         );
 
         require (
-            _preStateRootBatchHeader.prevTotalElements + _preStateRootProof.index == _transactionBatchHeader.prevTotalElements + _transactionProof.index,
+            _preStateRootBatchHeader.prevTotalElements + _preStateRootProof.index + 1 == _transactionBatchHeader.prevTotalElements + _transactionProof.index,
```
kfichter, line 125: `preStateRoot` for a given `transaction` sits at `transaction.index - 1`, not at `transaction.index`.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -173,81 +182,62 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
-            })
-        );
-    }
-
-    /**
-     * Allows a user to prove that an account does *not* exist in the state.
-     * @param _ovmContractAddress Address of the contract on the OVM.
-     * @param _stateTrieWitness Proof of the (empty) account state.
-     */
-    function proveEmptyContractState(
```
kfichter, line 213: Unnecessary function, removed.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -157,13 +167,12 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
      * Allows a user to prove the initial state of a contract.
      * @param _ovmContractAddress Address of the contract on the OVM.
-     * @param _account Claimed account state.
+     * @param _ethContractAddress Address of the corresponding contract on L1.
      * @param _stateTrieWitness Proof of the account state.
      */
     function proveContractState(
         address _ovmContractAddress,
         address _ethContractAddress,
-        Lib_OVMCodec.EVMAccount memory _account,
```
kfichter, line 166: No need to provide the account state, we already get it from the proof.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -173,81 +182,62 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
-            Lib_SecureMerkleTrie.verifyExclusionProof(
-                abi.encodePacked(_ovmContractAddress),
-                _stateTrieWitness,
-                preStateRoot
-            ),
-            "Account is not empty or invalid inclusion proof provided."
+        (
+            bool exists,
+            bytes memory encodedAccount
+        ) = Lib_SecureMerkleTrie.get(
```
kfichter, line 195: Function throws if the proof is not a valid inclusion or exclusion proof.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -173,81 +182,62 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
 
-        ovmStateManager.putEmptyAccount(_ovmContractAddress);
+        if (exists == true) {
+            Lib_OVMCodec.EVMAccount memory account = Lib_OVMCodec.decodeEVMAccount(
+                encodedAccount
+            );
+
+            address ethContractAddress = _ethContractAddress;
+            if (account.codeHash == EMPTY_ACCOUNT_CODE_HASH) {
+                ethContractAddress = 0x0000c0De0000C0DE0000c0de0000C0DE0000c0De;
```
kfichter, line 208: Use a known empty account to avoid an attack in which the user deploys code to the contract after supplying it here.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -266,33 +256,34 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
-        bytes memory value = Lib_RLPReader.readBytes(encodedValue);
+        bytes32 storageRoot = ovmStateManager.getAccountStorageRoot(_ovmContractAddress);
+        bytes32 value;
 
-        if (exists == true) {
-            require(
-                keccak256(value) == keccak256(abi.encodePacked(_value)),
-                "Provided storage slot value is invalid."
-            );
+        if (storageRoot == EMPTY_ACCOUNT_STORAGE_ROOT) {
```
kfichter, line 262: If the storage trie was empty, then we can only insert empty values.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -266,33 +256,34 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
+                bool exists,
+                bytes memory encodedValue
+            ) = Lib_SecureMerkleTrie.get(
+                abi.encodePacked(_key),
+                _storageTrieWitness,
+                storageRoot
             );
+
+            if (exists == true) {
+                value = Lib_BytesUtils.toBytes32PadLeft(
```
kfichter, line 275: Storage trie values are right aligned (padded with 0s on the left).

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -353,8 +344,13 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
         contributesToFraudProof(preStateRoot, transactionHash)
     {
         require(
+            ovmStateManager.getTotalUncommittedContractStorage() == 0,
```
kfichter, line 347: Force users to prove all storage slots first so that we know the `storageRoot` values are finalized.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -367,19 +363,21 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
             _stateTrieWitness,
             postStateRoot
         );
+
+        emit AccountCommitted(
```
kfichter, line 367: Need to emit an event to help clients figure out proof ordering.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -398,22 +396,18 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
         account.storageRoot = Lib_SecureMerkleTrie.update(
             abi.encodePacked(_key),
             Lib_RLPWriter.writeBytes(
-                abi.encodePacked(value)
+                Lib_Bytes32Utils.removeLeadingZeros(value)
```
kfichter, line 399: Ethereum strips all leading zero bytes from trie values before RLP encoding.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -398,22 +396,18 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
-            Lib_OVMCodec.encodeEVMAccount(
-                Lib_OVMCodec.toEVMAccount(account)
-            ),
-            _stateTrieWitness,
-            postStateRoot
-        );
-
         ovmStateManager.putAccount(_ovmContractAddress, account);
+
+        emit ContractStorageCommitted(
```
kfichter, line 407: Another event to help clients figure out ordering.

contracts/optimistic-ethereum/libraries/trie/Lib_MerkleTrie.sol
```diff
@@ -212,7 +210,7 @@ library Lib_MerkleTrie {
         )
     {
         return keccak256(_makeLeafNode(
-            _key,
+            Lib_BytesUtils.toNibbles(_key),
```
kfichter, line 213: `_makeLeafNode` expects nibbles as an input, this function was broken without first converting to nibbles.

kfichter (commented):
Left comments that might be useful

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
tynes: When are you planning on updating `_commitPendingAccount`?

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
tynes: Should some of these comments be in the code itself?

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
tynes: At one point this require was split up into many requires so that its easier to tell what the problem is, placing it all in a single require to save gas?

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
karlfloersch: 😭

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: The problem is in other codepaths, `_commitPendingAccount` needs to set with 1.  One option is to add it as an input, not sure if necessary

ben-chain: @kfichter actually, can we break this out into a separate PR?  @axchu is currently assigned to it, can reassign to you since solved here, but auditors want separate PR for this fix

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
kfichter: Hmmm that might've been a different `require` statement, this one was always in a single statement.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
kfichter: Sure thing, will remove the line from this PR and make a note.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
kfichter: Yeah, sorry. Am cleaning up this PR as I fix the tests.

contracts/optimistic-ethereum/libraries/trie/Lib_MerkleTrie.sol
```diff
@@ -145,9 +145,7 @@ library Lib_MerkleTrie {
     {
         // Special case when inserting the very first node.
         if (_root == KECCAK256_RLP_NULL_BYTES) {
-            return keccak256(
-                _makeLeafNode(_key, _value).encoded
-            );
+            return getSingleNodeRootHash(_key, _value);
```
kfichter, line 148: Refactor to use `getSingleNodeRootHash` since the logic was the same in both places.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: I also have some `run` order changes in [this PR](https://github.com/ethereum-optimism/contracts-v2/pull/94/files), but let's merge this in and I can update mine.

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
ben-chain, line 396: Note this function will return `false` if the `ITEM_STATE` is `ItemState.ITEM_COMMITTED`, not sure if intended behavior but would update either name or implementation to reflect.

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -578,11 +606,69 @@ contract OVM_StateManager is iOVM_StateManager {
         return totalUncommittedContractStorage;
     }
 
+    function wasContractStorageChanged(
```
ben-chain, line 609: Ditto for account comment above that this does not return `true` for `ITEM_COMMITTED`

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -578,11 +606,69 @@ contract OVM_StateManager is iOVM_StateManager {
...
+        bytes32 item = _getContractStorageItemHash(_contract, _key);
+        return itemStates[item] == ItemState.ITEM_COMMITTED;
+    }
+
 
     /**********************
      * Internal Functions *
      **********************/
 
+    function _getAccountItemHash(
```
ben-chain, line 644: nit: these could both be a `_getItemHash` with argument-dependent lookups, up to you.

contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
```diff
@@ -30,6 +30,7 @@ contract OVM_SequencerEntrypoint {
      * calldata[00:01]: transaction type (0 == EIP 155, 2 == Eth Sign Message)
      * calldata[01:33]: signature "r" parameter
      * calldata[33:65]: signature "s" parameter
+     * calldata[65:66]: signature "v" parameter
```
ben-chain, line 33: This is [a ticket](https://github.com/ethereum-optimism/roadmap/issues/273), can we actually not do this here so that there is a unique/standalone PR to give auditors? <3

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
ben-chain: Recommend to set this to ZERO_ADDRESS--less fun, but I fear this may be confusing, especially since we have a similar use of `c0dec` elsewhere IIRC.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -173,81 +182,69 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
 
-        ovmStateManager.putEmptyAccount(_ovmContractAddress);
+        if (exists == true) {
+            // Account exists, this was an inclusion proof.
+            Lib_OVMCodec.EVMAccount memory account = Lib_OVMCodec.decodeEVMAccount(
+                encodedAccount
+            );
+
+            address ethContractAddress = _ethContractAddress;
+            if (account.codeHash == EMPTY_ACCOUNT_CODE_HASH) {
```
ben-chain, line 209: I'm actually a little confused by this -- can you clarify the exact scenario that will trigger this?  Intuitively, I see no way for there to be an inclusion proof in the OVM state that has empty code at the bottom, since we do not have EOAs and do not have `ovmSELFDESTRUCT`.

If there's no scenario that will trigger this but we just thought of it, still could be worth leaving in as it was a great catch--moreso just curious.  Can initcode return 0 bytes maybe?

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -266,32 +263,38 @@ contract OVM_StateTransitioner is Lib_AddressResolver, OVM_FraudContributor, iOV
...
+            // Storage trie was empty, so the user is always allowed to insert zero-byte values.
+            value = bytes32(0);
         } else {
-            require(
-                _value == bytes32(0),
-                "Provided storage slot value is invalid."
+            (
+                bool exists,
+                bytes memory encodedValue
+            ) = Lib_SecureMerkleTrie.get(
```
ben-chain, line 276: recommend we duplicate the "this call will revert" explanation similar to the comment above.

contracts/optimistic-ethereum/libraries/utils/Lib_BytesUtils.sol
```diff
@@ -170,6 +170,20 @@ library Lib_BytesUtils {
...
+    function toBytes32PadLeft(
+        bytes memory _bytes
+    )
+        internal
+        pure
+        returns (bytes32)
+    {
+        bytes32 ret;
+        assembly {
+            ret := shr(mul(sub(32, mload(_bytes)), 8), mload(add(_bytes, 32)))
```
ben-chain, line 182: I am worried this will underflow if `_bytes.length > 32`.  Not sure how we want to handle this or what expected behavior should be if that condition is true--probably just truncate?

ben-chain (approved):
Left a few requests but LGTM -- hype!!!

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
kfichter: Great point, will change.

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
kfichter: Good point, changing.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
kfichter: I believe the only scenario that can trigger this is when initcode returns 0 bytes (just tested on remix, definitely possible).

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
kfichter: Adding.

contracts/optimistic-ethereum/libraries/utils/Lib_BytesUtils.sol
kfichter: Ah we should truncate. Will update.

contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
kfichter: lolol yes

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
kfichter: Fixed.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1185,6 +1191,11 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
     )
         internal
     {
+        // We don't set storage if the value didn't change.
+        if (_getContractStorage(_contract, _key) == _value) {
+            return;
+        }
+
```
ben-chain, line 1198: Is this an optimization or was it somehow a requirement?  If requirement, can you specify why in the comment?  Not obvious to me why this is being done

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1234,19 +1245,21 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-        // If we hadn't already changed the account, then we'll need to charge "nuisance gas" based
-        // on the size of the contract code.
+        // If we hadn't already changed the account, then we'll need to charge a constant
+        // "nuisance gas" that covers the cost of updating the account state.
         if (_wasAccountAlreadyChanged == false) {
             ovmStateManager.incrementTotalUncommittedAccounts();
-            _useNuisanceGas(
-                (Lib_EthUtils.getCodeSize(_getAccountEthAddress(_address)) * NUISANCE_GAS_PER_CONTRACT_BYTE) + MIN_NUISANCE_GAS_PER_CONTRACT
-            );
+            _useNuisanceGas(MIN_NUISANCE_GAS_PER_CONTRACT);
```
ben-chain, line 1262: Same thing here as with nonce fix -- this is audit L01, can we break into separate PR for convenience?  Happy to do it myself if you need it real quick.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1304,15 +1317,9 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-        if (gasleft() < MIN_GAS_FOR_INVALID_STATE_ACCESS) {
-            _revertWithFlag(RevertFlag.OUT_OF_GAS);
-        }
-
-        // See `_checkContractStorageLoad` for more information.
-        if (ovmStateManager.hasContractStorage(_contract, _key) == false) {
-            _revertWithFlag(RevertFlag.INVALID_STATE_ACCESS);
-        }
+        // Start by checking for load to make sure we have the storage slot and that we charge the
+        // "nuisance gas" necessary to prove the storage slot state.
```
ben-chain, line 1321:
```suggestion
        // Start by checking for load to make sure we have the storage slot in the state manager,
        // and that we charge the "nuisance gas" necessary to prove the storage slot state.
```

contracts/optimistic-ethereum/libraries/utils/Lib_BytesUtils.sol
```diff
@@ -170,6 +170,21 @@ library Lib_BytesUtils {
...
 
+    function toBytes32PadLeft(
+        bytes memory _bytes
+    )
+        internal
+        pure
+        returns (bytes32)
+    {
+        bytes32 ret;
+        uint256 len = _bytes.length <= 32 ? _bytes.length : 32;
```
ben-chain, line 181: 👍

ben-chain (approved):
Woohoo, LGTM!!

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
kfichter: Ah sorry yes -- this is a requirement. Reason is that if we change the storage slot, then we mark it as "changed" in the state manager and it has to be committed into the state trie. Normally this isn't a problem (because updating a value to the same value won't change the root) *except* in the case that the state trie was originally empty and a contract attempts to store `0x00.....`. This causes a discrepancy because the default storage root is `keccak256('80')` but once you store a value (even if it's `0x00...`), that storage root changes. Will reflect in a comment.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
kfichter: Ah ok, will revert the change.

PR #99, opened 2020-12-08
Make deployments consistently succeed  

karlfloersch:
Currently deployments fail often, this PR makes them slower but more consistent.

## Contributing Agreement

### DIFF
--- a/src/contract-deployment/config.ts
+++ b/src/contract-deployment/config.ts
@@ -29,7 +29,7 @@ export interface RollupDeployConfig {
     owner: string | Signer
     allowArbitraryContractDeployment: boolean
   }
-  deployOverrides?: Overrides
+  deployOverrides: Overrides
   dependencies?: string[]
 }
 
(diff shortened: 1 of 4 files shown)


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -355,7 +355,7 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
 
         for (uint32 i = 0; i < numContexts; i++) {
             BatchContext memory context = _getBatchContext(i);
-            _validateBatchContext(context, nextQueueIndex);
+            // _validateBatchContext(context, nextQueueIndex);
```
tynes, line 358: Did you mean to comment this out?

src/contract-deployment/deploy.ts
```diff
@@ -16,15 +16,18 @@ export interface DeployResult {
 export const deploy = async (
   config: RollupDeployConfig
 ): Promise<DeployResult> => {
+  console.log('Beginning deploy')
```
tynes, line 19: These logs will break the `deployer` container

src/contract-deployment/config.ts
```diff
@@ -65,11 +65,12 @@ export const makeContractDeployConfig = async (
         )
           .connect(config.deploymentSigner)
           .attach(contracts.Proxy__OVM_L1CrossDomainMessenger.address)
-        await xDomainMessenger.initialize(AddressManager.address)
-        await AddressManager.setAddress(
+        await (await xDomainMessenger.initialize(AddressManager.address, config.deployOverrides)).wait()
```
tynes, line 68: Not a huge fan of nested awaits on the same line, it feels pretty busy

tynes:
See this approach for a faster approach: https://github.com/ethereum-optimism/contracts-v2/pull/51/files#diff-3d2aa7498c1a7bdecb7dfd7b4b32db4b9002a9dd0093316455609da76ff952f1R60

karlfloersch:
> See this approach for a faster approach: https://github.com/ethereum-optimism/contracts-v2/pull/51/files#diff-3d2aa7498c1a7bdecb7dfd7b4b32db4b9002a9dd0093316455609da76ff952f1R60

Unfortunately on Goerli sending more than 1 transaction without waiting for the receipt often causes the deployment to fail :( So that should work if mempool management was better but unfortunately we've had issues in the past.

karlfloersch:
Closing in favor of https://github.com/ethereum-optimism/contracts-v2/pull/99

PR #100, opened 2020-12-08
Add queue migraiton to deploy

karlfloersch:
## Description
WIP:

TODO:
- [ ] Add simple queue migration to the normal deploy script
- [ ] Make it a separate script so we don't always run it.

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -257,6 +257,47 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         );
     }
 
+    function authenticatedEnqueue(
+        address _msgSender,
+        address _target,
+        uint256 _gasLimit,
+        bytes memory _data,
+        uint256 _theTimestamp,
+        uint256 _theBlockNumber
+    )
+        public
+    {
+        bytes32 transactionHash = keccak256(
+            abi.encode(
+                _msgSender,
+                _target,
+                _gasLimit,
+                _data
+            )
+        );
+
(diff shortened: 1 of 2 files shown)


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -257,6 +257,47 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
 
+    function authenticatedEnqueue(
+        address _msgSender,
+        address _target,
+        uint256 _gasLimit,
+        bytes memory _data,
+        uint256 _theTimestamp,
+        uint256 _theBlockNumber
+    )
+        public
```
tynes, line 268: Is this supposed to be authenticated?

src/contract-deployment/config.ts
```diff
@@ -87,6 +89,50 @@ export const makeContractDeployConfig = async (
...
+        const otherAddressManager = getContractFactory('Lib_AddressManager').attach(otherAddressManagerAddress).connect(otherSigner)
+
+        // ~~~~ init addrs ~~~~
+        const otherCtcAddress = await otherAddressManager.getAddress('OVM_CanonicalTransactionChain')
+        const otherCtc = getContractFactory('OVM_CanonicalTransactionChain').attach(otherCtcAddress).connect(otherSigner)
+
+        // ~~~~ filter ~~~~
+        const filter = {
+          address: otherCtc.address,
+          topics: [ethers.utils.id(`TransactionEnqueued(address,address,uint256,bytes,uint256,uint256)`)],
```
tynes, line 111: I believe a helper function exists on `otherCtc` for each event, so you can call `otherCtc.TransactionEnqueued()` which will return the same thing as `ethers.utils.id` and is more safe

PR #103, opened 2020-12-08
deploy targets: add configurable deployment targets

tynes:
## Description

Adds the ability to specify a deployment target and the ability to selectively deploy to either L1 or L2

## Contributing Agreement

### DIFF
--- a/src/contract-deployment/deploy.ts
+++ b/src/contract-deployment/deploy.ts
@@ -39,7 +39,12 @@ export const deploy = async (
       continue
     }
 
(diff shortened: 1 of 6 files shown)


src/contract-deployment/deploy.ts
```diff
@@ -38,6 +39,15 @@ export const deploy = async (
       continue
     }
```
tynes, line 41: The linter makes this very difficult to read

ben-chain:
@tynes do you think this is still relevant?  I can clean this up if so.

smartcontracts:
`hardhat-deploy` provides pretty good native support for this (we can attach custom tags to each contract).

PR #106, opened 2020-12-08
global root index check

axchu:
### Fixes
- Fixes # https://github.com/ethereum-optimism/roadmap/issues/257

## Contributing Agreement


ben-chain (approved):
Great, LGTM!  We may have uncovered an off-by-1 error where it should be `-1` instead of `+1`, but let's merge this and @kfichter can incorporate into his PR if needed, as it is a separate fix from this.

PR #121, opened 2020-12-11
OZ L09 Remove unnecessary use of assembly

axchu:
### Fixes
- Fixes https://github.com/ethereum-optimism/roadmap/issues/264

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/libraries/rlp/Lib_RLPWriter.sol
+++ b/contracts/optimistic-ethereum/libraries/rlp/Lib_RLPWriter.sol
@@ -297,22 +297,4 @@ library Lib_RLPWriter {
 
         return flattened;
     }
-
-    /**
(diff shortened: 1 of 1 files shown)


contracts/optimistic-ethereum/libraries/rlp/Lib_RLPWriter.sol
```diff
@@ -89,16 +89,7 @@ library Lib_RLPWriter {
             bytes memory _out
         )
     {
-        _malloc(0x20);
```
ben-chain, line 92: I think that, with this change, `_malloc` is no longer used -- can you check and delete if true?

ben-chain (approved):
LGTM!  Noted one potential removal I think you can do; good to go once you've checked that.

PR #137, opened 2020-12-13
Fix rlp encoding bug in contract dump

K-Ho:
## Description
Instead of RLP decoding values from the storage trie, we were just slicing off the first byte. This works for most values, but values between `0x00` and 0x7f` are not prefixed when rlp encoding, so these values were getting wiped. 
## Questions
- 
-
-

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement


karlfloersch (approved):
LGTM!

PR #139, opened 2020-12-13
Match transaction parameter ordering

axchu:
### Fixes
- Fixes https://github.com/ethereum-optimism/roadmap/issues/276

## Contributing Agreement


axchu:
This will have downstream effects. Probably shouldn't merge until those are investigated.

ben-chain:
Yep, I think this breaks the batch submitter and potentially the provider (cc @tynes , assuming it would but not positive)

PR #150, opened 2020-12-16
ctc: disable appendQueueBatch

tynes:
## Description

Disables `appendQueueBatch` at the contract level with a `revert`. Needs to skip a few tests, adds a test for the newly added revert

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -266,6 +266,9 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         override
         public
     {
+        // Disable `appendQueueBatch` for minnet
+        revert("Cannot appendQueueBatch.");
+
         _numQueuedTransactions = Math.min(_numQueuedTransactions, getNumPendingQueueElements());
         require(
             _numQueuedTransactions > 0,
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -267,7 +267,7 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         public
     {
         // Disable `appendQueueBatch` for minnet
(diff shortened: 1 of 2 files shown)


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -266,6 +266,9 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         override
         public
     {
+        // Disable `appendQueueBatch` for minnet
+        revert("Cannot appendQueueBatch.");
```
ben-chain, line 270:
```suggestion
        revert("appendQueueBatch is currently disabled.");
```

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.spec.ts
```diff
@@ -500,7 +508,7 @@ describe('OVM_CanonicalTransactionChain', () => {
   })
 
   describe('verifyTransaction', () => {
-    it('should successfully verify against a valid queue transaction', async () => {
+    it.skip('should successfully verify against a valid queue transaction', async () => {
```
ben-chain, line 511: I actually think it's important to keep this test since the code path being tested has not been disabled, just the way we were getting there.  Recommend to replace line 523 with a `sequencerBatchAppend` and un-skipping.

ben-chain (changes requested):
Requested a change--I think important to keep one of the skipped tests, we can still hit the code path even with this contract change.

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.spec.ts
```diff
@@ -500,7 +508,66 @@ describe('OVM_CanonicalTransactionChain', () => {
...
+        ],
+        transactions: [],
+      })
+
+      expect(
+        await OVM_CanonicalTransactionChain.verifyTransaction(
+          {
+            timestamp,
+            blockNumber,
+            l1QueueOrigin: 1,
```
tynes, line 542: Nit: no enum being used here, also I thought there is only 1 queue on L1

tynes (commented):
This looks good to me, I cannot approve because I opened the PR

K-Ho (approved):
great, lgtm!

ben-chain (approved):

PR #163, opened 2020-12-20
Code freeze audit integrations

karlfloersch:
## Description
All of our audit fixes! 

## Contributing Agreement


test/helpers/constants.ts
```diff
@@ -17,6 +17,7 @@ export const DEFAULT_ACCOUNTS_BUIDLER = defaultAccounts.map((account) => {
 export const OVM_TX_GAS_LIMIT = 10_000_000
 export const RUN_OVM_TEST_GAS = 20_000_000
 export const FORCE_INCLUSION_PERIOD_SECONDS = 600
+export const FORCE_INCLUSION_PERIOD_BLOCKS = 7200
```
karlfloersch, line 20: Why is this 7200 vs 600 for the force inclusion period seconds?

karlfloersch: Ah I think this is the opposite of what it should be -- 600/12 not 600*12

karlfloersch (commented):
LGTM!

I can't approve because I just so happened to have created this PR but I didn't write it so I feel confident in approving / shipping

PR #164, opened 2020-12-21
state dump: allow for config

tynes:
## Description

Enable config for state dump parameters

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

### DIFF
--- a/bin/take-dump.ts
+++ b/bin/take-dump.ts
@@ -3,13 +3,25 @@ import * as fs from 'fs'
 import * as path from 'path'
 import * as mkdirp from 'mkdirp'
 
+
+const env = process.env
+const CHAIN_ID = env.CHAIN_ID || '420'
+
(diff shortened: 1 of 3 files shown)


bin/take-dump.ts
```diff
@@ -3,13 +3,25 @@ import * as fs from 'fs'
 import * as path from 'path'
 import * as mkdirp from 'mkdirp'
 
+
+const env = process.env
+const CHAIN_ID = env.CHAIN_ID || '420'
```
K-Ho, line 8: Can we make this environment var mandatory? Think it would save us a lot of headache to force everyone to explicitly specify their chainID when generating a dump.

K-Ho: resolved, think it's ok after all

K-Ho (approved):
lgtm!! tested locally

PR #167, opened 2021-01-06
deploy: fix relayer address config

tynes:
## Description

Fixes an issue around relayer authentication where the relayer address can now be configured and is set to the zero address by default, meaning there is no withdrawal authentication

## Contributing Agreement

### DIFF
--- a/bin/deploy.js
+++ b/bin/deploy.js
@@ -25,13 +25,14 @@ const HD_PATH = env.HD_PATH || utils.defaultPath;
 const BLOCK_TIME_SECONDS = env.BLOCK_TIME_SECONDS || 15;
 const L2_CROSS_DOMAIN_MESSENGER_ADDRESS =
   env.L2_CROSS_DOMAIN_MESSENGER_ADDRESS || '0x4200000000000000000000000000000000000007';
-
-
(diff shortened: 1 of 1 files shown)


bin/deploy.js
```diff
@@ -69,7 +80,7 @@ const L2_CROSS_DOMAIN_MESSENGER_ADDRESS =
       L2CrossDomainMessengerAddress: L2_CROSS_DOMAIN_MESSENGER_ADDRESS
     },
     l1CrossDomainMessengerConfig: {
-      relayerAddress: SEQUENCER_ADDRESS,
+      relayerAddress: RELAYER_ADDRESS,
```
karlfloersch, line 83: Ok great that seems like a better option than what we had before.

K-Ho (approved):
awesome lgtm! So this will not require authentication currently in the integrations repo because there is no `RELAYER_PRIVATE_KEY` set, right?

tynes:
> awesome lgtm! So this will not require authentication currently in the integrations repo because there is no `RELAYER_PRIVATE_KEY` set, right?

Exactly, it is something that we could add in the future by setting the correct env var in the optimism integration repo

karlfloersch (approved):

PR #169, opened 2021-01-07
Unskip tests that were waiting for an update to smock

maurelian:
## Description

I found some tests that were skipped, that can/should be re-activated.

This is a tiny PR, so seemed worth doing now.


karlfloersch (approved):
LGTM considering the tests pass! Woot!

PR #105, opened 2020-12-08
Audit/tob 6 message passing reentrancy risk 201

axchu:
### Fixes
- Fixes # https://github.com/ethereum-optimism/roadmap/issues/201

## Contributing Agreement

- [x] I have read and understood the [Optimism Contributing Guide and Code of Conduct](https://github.com/ethereum-optimism/optimism-monorepo/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

PR #172, opened 2021-01-09
Add nonReentrant to `relayMessage()`

maurelian:
## Description

This is a fix to Sam's bug. 

## Questions

> Additionally, @karlfloersch pointed out that another way to address this would be checking that the msg.sender == address(0), because this condition can only be satisfied at the first contract ovmCALLed by run.

### Fixes
- Fixes #105 
- Also https://github.com/ethereum-optimism/roadmap/issues/394 (private issue)

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_BaseCrossDomainMessenger.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_BaseCrossDomainMessenger.sol
@@ -40,9 +40,10 @@ abstract contract OVM_BaseCrossDomainMessenger is iOVM_BaseCrossDomainMessenger,
         address _target,
         bytes memory _message,
         uint32 _gasLimit
-    ) nonReentrant
+    )
         override
         public
+        nonReentrant
     {
         bytes memory xDomainCalldata = _getXDomainCalldata(
             _target,
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_BaseCrossDomainMessenger.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_BaseCrossDomainMessenger.sol
@@ -43,7 +43,6 @@ abstract contract OVM_BaseCrossDomainMessenger is iOVM_BaseCrossDomainMessenger,
     )
         override
         public
-        nonReentrant
     {
         bytes memory xDomainCalldata = _getXDomainCalldata(
             _target,
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
@@ -7,6 +7,7 @@ import { Lib_OVMCodec } from "../../libraries/codec/Lib_OVMCodec.sol";
 import { Lib_AddressResolver } from "../../libraries/resolver/Lib_AddressResolver.sol";
 import { Lib_AddressManager } from "../../libraries/resolver/Lib_AddressManager.sol";
 import { Lib_SecureMerkleTrie } from "../../libraries/trie/Lib_SecureMerkleTrie.sol";
+import { Lib_ReentrancyGuard } from "../../libraries/utils/Lib_ReentrancyGuard.sol";
 
 /* Interface Imports */
 import { iOVM_L1CrossDomainMessenger } from "../../iOVM/bridge/iOVM_L1CrossDomainMessenger.sol";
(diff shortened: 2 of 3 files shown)


ben-chain (dismissed):
Lookin' good!  I see the question above -- to clarify, this will only work in the `OVM_L2CrossDomainMessenger`.  The reason is because of account abstraction,[ OVM state transitions](https://github.com/ethereum-optimism/contracts-v2/blob/25a2349b7839411677e397b507964a4e45497ee6/contracts/optimistic-ethereum/libraries/codec/Lib_OVMCodec.sol#L78) do not include a `from`, just a `to`, meaning that the `entrypoint` contract will get `address(0)` for `ovmCALLER`, during the [very first `ovmCALL`](https://github.com/ethereum-optimism/contracts-v2/blob/25a2349b7839411677e397b507964a4e45497ee6/contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol#L177) only.  Since the contract uses the compiler (`+build ovm`), we can easily access `ovmCALLER` with a normal `msg.sender`.

It strikes me that this property could be worth explicitly adding to our Execution Manager tests which use the `ExecutionManagerTestRunner`.  I think we want to merge this ASAP though, so I'm gonna approve this one and create another ticket for that.

contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
```diff
@@ -18,6 +19,7 @@ import { OVM_BaseCrossDomainMessenger } from "./OVM_BaseCrossDomainMessenger.sol
 
 /**
  * @title OVM_L1CrossDomainMessenger
+ * @dev This contract lives on L1. It sends messages to L2, and relays them from L2
```
ben-chain, line 22: nit suggestion:

```suggestion
 * @dev This contract lives on L1. It sends L1->L2 messages into L2, and relays L2->L1 messages from L2 to their target on L1.
```

contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
```diff
@@ -140,6 +143,7 @@ contract OVM_L1CrossDomainMessenger is iOVM_L1CrossDomainMessenger, OVM_BaseCros
     )
         override
         public
+        nonReentrant
```
ben-chain, line 146: FWIW, this code path speaks directly to the rollup contracts (`OVM_CanonicalTransactionChain.enqueue` specifically) so it should not be re-enterable based on the behavior of that target, so it should not be needed.

I'd recommend removing it, because the L1 `SSTORE` is gonna be expensive.  If you think that it's worth keeping, we can, but we should also add it to `sendMessage`, since the functions serve similar purposes.

contracts/optimistic-ethereum/OVM/bridge/OVM_L2CrossDomainMessenger.sol
```diff
@@ -17,6 +18,7 @@ import { OVM_BaseCrossDomainMessenger } from "./OVM_BaseCrossDomainMessenger.sol
 /**
  * @title OVM_L2CrossDomainMessenger
  * @dev L2 CONTRACT (COMPILED)
+ * This contract lives on L2. It sends messages to L1, and relays them from L1
```
ben-chain, line 21:
```suggestion
 * This contract lives on L2. It sends messages to L1, and relays them from L1.
```

ben-chain (changes requested):
Requested a couple small changes but lookin' good!

contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
maurelian: `OVM_BaseCrossDomainMessenger.sendMessage()` already has the `nonReentrant` modifier on it. You can see above that I moved it down a couple lines to improve the style. But looking at both `sendMessage` and `replayMessage` more  closely, I agree that it looks pretty safe to remove them.

contracts/optimistic-ethereum/OVM/bridge/OVM_BaseCrossDomainMessenger.sol
```diff
@@ -40,7 +40,7 @@ abstract contract OVM_BaseCrossDomainMessenger is iOVM_BaseCrossDomainMessenger,
         address _target,
         bytes memory _message,
         uint32 _gasLimit
-    ) nonReentrant
+    )
```
maurelian, line 43: OK, I've removed `nonReentrant` from `sendMessage` and `replayMessage`. This _feels_ safe because on L1 those functions both call only `OVM_CanonicalTransactionChain`, and on L2 we only have `sendMessage` which calls to the `OVM_L2ToL1MessagePasser`.

ben-chain (approved):
Beautiful, LGTM!

PR #173, opened 2021-01-11
weth: don't emit events

tynes:
## Description

It was observed that the Wrapped Eth precompile contract emits events for fee transfers. This means an additional event will happen for each transaction. @karlfloersch mentioned that this should be removed and it was a really fast change so here is the PR.

I'm 50/50 on thinking this is a good idea - while layer 1 does not have an event emitted for the fee being taken, I think that it could be easily explainable to developers since it is an erc20 being used. The approach that I take in this PR removes the events completely, which also removes events for non tx fee based erc20 transfers. I feel like there should be events for these kinds of transfers. A way to solve this would be to conditionally emit the events, but I think that is kind of strange because events are nice to be able to recreate the state of the contract without needing to play all of the transactions. This is a long way of saying that I don't think this change should be merged without more discussion

## Contributing Agreement


tynes:
Closing this due to the desire to keep the events around for WETH

PR #174, opened 2021-01-11
ctc: public getters, remove dead variable

tynes:
## Description

Removes unused `lastOVMTimestamp` and makes the constructor args public variables so that they can be fetched by users when constructing transactions. Also useful for the sequencer to configure its gas limit on L2

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -42,10 +42,9 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
      * Variables *
      *************/
 
-    uint256 internal forceInclusionPeriodSeconds;
-    uint256 internal forceInclusionPeriodBlocks;
-    uint256 internal maxTransactionGasLimit;
-    uint256 internal lastOVMTimestamp;
+    uint256 public forceInclusionPeriodSeconds;
+    uint256 public forceInclusionPeriodBlocks;
+    uint256 public maxTransactionGasLimit;
 
 
     /***************


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -42,10 +42,9 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
      * Variables *
      *************/
 
-    uint256 internal forceInclusionPeriodSeconds;
-    uint256 internal forceInclusionPeriodBlocks;
-    uint256 internal maxTransactionGasLimit;
-    uint256 internal lastOVMTimestamp;
+    uint256 public forceInclusionPeriodSeconds;
+    uint256 public forceInclusionPeriodBlocks;
+    uint256 public maxTransactionGasLimit;
```
ben-chain, line 47: Technically the gas limit here is only one of two duplicate checks, the other one being [here](https://github.com/ethereum-optimism/contracts-v2/blob/4e115b68c4bf30589c0d4329ffab6cf56f9fbffb/contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol#L1622).  This check in the CTC is used on L1 during `enqueue` for deposits, but NOT checked against sequencer transactions, so the one I linked in the execution manager may be better to set geth based on, until we resolve everything so that there's only the single source of truth--tech debt we need to bail out!

ben-chain (approved):
LGTM!  There is one caveat I left in a comment worth mentioning, though.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
tynes: Thanks for the heads up there!

PR #176, opened 2021-01-12
Final Parameterization of Constants

ben-chain:
## Description
Updates the following values:
- `MAX_TRANSACTION_GAS_LIMIT`
- `forceInclusionPeriodSeconds` and `forceInclusionPeriodBlocks`
- `MIN_ROLLUP_TX_GAS`
- `L2_GAS_DISCOUNT_DIVISOR`
- `chainId`

### Fixes
- Fixes https://github.com/ethereum-optimism/roadmap/issues/305

## Contributing Agreement


bin/deploy.js
```diff
@@ -11,15 +11,15 @@ const sequencerKey = env.SEQUENCER_PRIVATE_KEY;
...
 const MAX_GAS_PER_QUEUE_PER_EPOCH = env.MAX_GAS_PER_QUEUE_PER_EPOCH || 250000000;
 const SECONDS_PER_EPOCH = env.SECONDS_PER_EPOCH || 0;
 let WHITELIST_OWNER = env.WHITELIST_OWNER;
 const WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT = env.WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT || true;
-const FORCE_INCLUSION_PERIOD_SECONDS = env.FORCE_INCLUSION_PERIOD_SECONDS || (60 * 300000); // 30 min
+const FORCE_INCLUSION_PERIOD_SECONDS = env.FORCE_INCLUSION_PERIOD_SECONDS || ethers.constants.MaxUint256;
 const FRAUD_PROOF_WINDOW_SECONDS = env.FRAUD_PROOF_WINDOW_SECONDS || (60 * 60 * 24 * 7); // 7 days
 const SEQUENCER_PUBLISH_WINDOW_SECONDS = env.SEQUENCER_PUBLISH_WINDOW_SECONDS || (60 * 30); // 30 min
-const CHAIN_ID = env.CHAIN_ID || 420; // layer 2 chainid
+const CHAIN_ID = env.CHAIN_ID || 422; // layer 2 chainid
```
ben-chain, line 22: @tynes this will require updates elsewhere in the codebase, I don't know the exhaustive list.

bin/deploy.js
```diff
@@ -11,15 +11,15 @@ const sequencerKey = env.SEQUENCER_PRIVATE_KEY;
 let SEQUENCER_ADDRESS = env.SEQUENCER_ADDRESS;
 const web3Url = env.L1_NODE_WEB3_URL || 'http://127.0.0.1:8545';
 const MIN_TRANSACTION_GAS_LIMIT = env.MIN_TRANSACTION_GAS_LIMIT || 50000;
-const MAX_TRANSACTION_GAS_LIMIT = env.MAX_TRANSACTION_GAS_LIMIT || Math.floor(9000000 * 63/64 - 100000); // = 8_759_375, Default max for Kovan
+const MAX_TRANSACTION_GAS_LIMIT = env.MAX_TRANSACTION_GAS_LIMIT || 9000000;
```
ben-chain, line 14: This is also related to the geth hanging bug represented elsewhere

bin/deploy.js
tynes: I believe for geth, a CLI flag can dynamically set the chain id so we are good there

tynes:
I think its ok for now to hard code in these params, but I think in the longer term it'll be easier to import these kinds of things from a JSON file so that we can maintain params for multiple deployments and then we can pass the network to the deploy script and pull the config params from the JSON file

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -26,9 +26,9 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
      *************/
 
     // L2 tx gas-related
-    uint256 constant public MIN_ROLLUP_TX_GAS = 20000;
+    uint256 constant public MIN_ROLLUP_TX_GAS = 100000;
     uint256 constant public MAX_ROLLUP_TX_SIZE = 10000;
```
tynes, line 30: Does making these constants save a lot of gas compared to making them constructor args? Its just going to be difficult to maintain having to make a commit if we want to change these params between different deployments. Then we will need to call them to remember what they are for the different deployments, unless we maintain a bunch of different git branches for each deployment which seems like a lot of extra work

src/contract-dumps.ts
```diff
@@ -114,12 +114,12 @@ export const makeStateDump = async (cfg: RollupDeployConfig): Promise<any> => {
     deploymentSigner: signer,
     ovmGasMeteringConfig: {
       minTransactionGasLimit: 0,
-      maxTransactionGasLimit: 1_000_000_000,
+      maxTransactionGasLimit: 9_000_000,
```
karlfloersch, line 117: Ah this config option is I believe just for generating the state dump & that makes me lean towards keeping it as an insane value

bin/deploy.js
```diff
@@ -11,15 +11,15 @@ const sequencerKey = env.SEQUENCER_PRIVATE_KEY;
...
 const MAX_GAS_PER_QUEUE_PER_EPOCH = env.MAX_GAS_PER_QUEUE_PER_EPOCH || 250000000;
 const SECONDS_PER_EPOCH = env.SECONDS_PER_EPOCH || 0;
 let WHITELIST_OWNER = env.WHITELIST_OWNER;
 const WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT = env.WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT || true;
-const FORCE_INCLUSION_PERIOD_SECONDS = env.FORCE_INCLUSION_PERIOD_SECONDS || (60 * 300000); // 30 min
+const FORCE_INCLUSION_PERIOD_SECONDS = env.FORCE_INCLUSION_PERIOD_SECONDS || ethers.constants.MaxUint256;
```
karlfloersch, line 19: Would reduce to fit in a JS number

bin/deploy.js
tynes: I believe this issue was fixed by https://github.com/ethereum-optimism/go-ethereum/commit/42c15d285804ce4bd77309dfdb1842c860d5c0f1

bin/deploy.js
```diff
@@ -11,15 +11,15 @@ const sequencerKey = env.SEQUENCER_PRIVATE_KEY;
...
 const web3Url = env.L1_NODE_WEB3_URL || 'http://127.0.0.1:8545';
 const MIN_TRANSACTION_GAS_LIMIT = env.MIN_TRANSACTION_GAS_LIMIT || 50000;
-const MAX_TRANSACTION_GAS_LIMIT = env.MAX_TRANSACTION_GAS_LIMIT || Math.floor(9000000 * 63/64 - 100000); // = 8_759_375, Default max for Kovan
+const MAX_TRANSACTION_GAS_LIMIT = env.MAX_TRANSACTION_GAS_LIMIT || 9000000;
 const MAX_GAS_PER_QUEUE_PER_EPOCH = env.MAX_GAS_PER_QUEUE_PER_EPOCH || 250000000;
 const SECONDS_PER_EPOCH = env.SECONDS_PER_EPOCH || 0;
 let WHITELIST_OWNER = env.WHITELIST_OWNER;
 const WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT = env.WHITELIST_ALLOW_ARBITRARY_CONTRACT_DEPLOYMENT || true;
-const FORCE_INCLUSION_PERIOD_SECONDS = env.FORCE_INCLUSION_PERIOD_SECONDS || (60 * 300000); // 30 min
+const FORCE_INCLUSION_PERIOD_SECONDS = env.FORCE_INCLUSION_PERIOD_SECONDS || 2592000; // 30 days
```
tynes, line 19: Double checked this number matches

tynes (approved):
This looks good to me

PR #196, opened 2021-01-17
Lib_RingBuffer.RingBuffer.nextOverwritableIndex

bruno-latour:
when caller has pushed one element in this buffer, the nextOverwritableIndex should be inc 1.

## Questions
- 
-
-

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/libraries/utils/Lib_RingBuffer.sol
+++ b/contracts/optimistic-ethereum/libraries/utils/Lib_RingBuffer.sol
@@ -84,7 +84,8 @@ library Lib_RingBuffer {
         // Index to write to is the difference of the global and reset indices.
         uint256 writeHead = ctx.globalIndex - ctx.currResetIndex;
         currBuffer.buf[writeHead] = _value;
-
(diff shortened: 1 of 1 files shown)


bruno-latour:
I want to review this repo code , thanks.

ben-chain:
@bruno-latour Thanks so much for taking a look around our code, so cool to see! 😄

The reason that we do not implement updates to the `lastOverwritableIndex` in this way is because it would make the total number of elements constant before getting overwritten.  However, in our model, what we want is for elements to only be overwritable *after a certain amount of time has passed*.  This may mean that in one time period, `push` gets called so many times that we have to use new storage slots.

Because of this, we opted for allowing an authenticated contract to set the field instead: [example usage](https://github.com/ethereum-optimism/contracts-v2/blob/0ad4dcfdef11ef87e278a8159de8414c8e329ba1/contracts/optimistic-ethereum/OVM/chain/OVM_ChainStorageContainer.sol#L213).  Hopefully this makes sense!

PR #197, opened 2021-01-20
tests: buidler gas report

tynes:
## Description

Adds an autogenerated gas report at the end of `yarn test`.
This was super simple to add, literally just importing something into the buidler config.
Maybe we want something different? It would be nice to have a way to know changes in the amount of gas used for any change in the contracts. This won't give that exactly, but is a step in the right direction.

## Contributing Agreement

### DIFF
--- a/buidler.config.ts
+++ b/buidler.config.ts
@@ -1,5 +1,7 @@
 import { usePlugin, BuidlerConfig } from '@nomiclabs/buidler/config'
 
+usePlugin('buidler-gas-reporter')
+
 import {
   DEFAULT_ACCOUNTS_BUIDLER,
   RUN_OVM_TEST_GAS,
@@ -11,7 +13,7 @@ usePlugin('buidler-typechain')
 
 import './plugins/buidler/ovm-compiler'
 
-const config: BuidlerConfig = {
+const config = {
   networks: {
     buidlerevm: {
       accounts: DEFAULT_ACCOUNTS_BUIDLER,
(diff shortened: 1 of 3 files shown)


buidler.config.ts
```diff
@@ -11,7 +13,7 @@ usePlugin('buidler-typechain')
 
 import './plugins/buidler/ovm-compiler'
 
-const config: BuidlerConfig = {
+const config = {
```
tynes, line 16: Had to remove the type from this as typescript would yell

tynes:
This does add quite a few dependencies, not sure if worth it. Thoughts?

buidler.config.ts
```diff
@@ -29,6 +31,9 @@ const config: BuidlerConfig = {
     outDir: 'build/types',
     target: 'ethers-v5',
   },
+  gasReporter: {
+    currency: 'USD',
```
ben-chain, line 35: pricing this in non-gas units seems like a very strange choice.

ben-chain:
Can this tool give us easy ways to add explicit assertions in tests about gas?  I think this is what we really need; to catch regressions in CI/testing.

buidler.config.ts
TransmissionsDev: it shows both gas units and usd price :p

gakonst:
As I wrote in chat, this should be prob closed in favor of gas snapshots?

PR #200, opened 2021-01-22
Add descriptive comments above the contract declaration for all 'non-abstract contracts'

maurelian:
## Description

We want to make life better for people reading the code, so we're adding a detailed comment at the top of each concrete contract.

This is WIP, but I'm looking for feedback on the approach. Some contracts just have an `@todo` comment so I can find them later. 

**Request:** I've used a different ordering and layout of info in different places, it would help to get feedback on which approach is best.

Some examples:

- https://github.com/ethereum-optimism/contracts-v2/blob/42497aa6775c43b0cea1cbd947a733b563eac1b5/contracts/optimistic-ethereum/OVM/chain/OVM_StateCommitmentChain.sol#L20-L32

- https://github.com/ethereum-optimism/contracts-v2/blob/42497aa6775c43b0cea1cbd947a733b563eac1b5/contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol#L25-L38

- https://github.com/ethereum-optimism/contracts-v2/blob/42497aa6775c43b0cea1cbd947a733b563eac1b5/contracts/optimistic-ethereum/OVM/execution/OVM_SafetyChecker.sol#L7-L17


maurelian:
Some specific questions: 
- which order is more helpful, long paragraph about functionality, then compiler and layer1/2 info or the otehr way around?
- should I say “compiles to EVM/OVM” or “Is compiled by the standard solc/our forked solc”
- should I write “The State Manager…” or “The OVM_StateManager”….

contracts/optimistic-ethereum/OVM/execution/OVM_SafetyChecker.sol
```diff
@@ -6,6 +6,14 @@ import { iOVM_SafetyChecker } from "../../iOVM/execution/iOVM_SafetyChecker.sol"
 
 /**
  * @title OVM_SafetyChecker
+ * @dev  The Safety Checker ensures that contracts deployed on L2 do not contain any
+ * "unsafe" opcodes. An opcode is considered unsafe if it would access state variables which
+ * are specific to the environment (ie. L1 or L2) in which it is executed, as this could be used
+ * to "escape the sandbox" of the OVM, resulting in non-deterministic fraud proofs. 
+ * That is, an attacker wouuld be able to "prove fraud" on an honestly applied transaction.
+ *
+ * This contract compiles to EVM bytecode.
```
ben-chain, line 15: One potential sticking point is that technically, OVM code is a subset of EVM bytecode.  Further, we have cases where some L2 contracts use the L1 compiler, and others use our fork.  I lean towards being overly verbose and specifying all of this.

I also like having this be at the top of the docstrings.  If we wanted to avoid them getting too long, I guess it could be 1 line, something like 
```
* Compiler used: {EVM,OVM,BOTH}.  Deploy target: {EVM,OVM,BOTH}.
```

contracts/optimistic-ethereum/OVM/execution/OVM_SafetyChecker.sol
```diff
@@ -6,6 +6,14 @@ import { iOVM_SafetyChecker } from "../../iOVM/execution/iOVM_SafetyChecker.sol"
 
 /**
  * @title OVM_SafetyChecker
+ * @dev  The Safety Checker ensures that contracts deployed on L2 do not contain any
```
ben-chain, line 9:
```suggestion
 * @dev  The Safety Checker verifies that contracts deployed on L2 do not contain any
```

contracts/optimistic-ethereum/OVM/execution/OVM_SafetyChecker.sol
```diff
@@ -6,6 +6,14 @@ import { iOVM_SafetyChecker } from "../../iOVM/execution/iOVM_SafetyChecker.sol"
...
+ * "unsafe" opcodes. An opcode is considered unsafe if it would access state variables which
```
ben-chain, line 10:
```suggestion
 * "unsafe" operations. An operation is considered unsafe if it would access state variables which
```

contracts/optimistic-ethereum/OVM/execution/OVM_SafetyChecker.sol
maurelian: I kind of addressed that OVM < EVM subset idea here: https://github.com/ethereum-optimism/contracts-v2/blob/42497aa6775c43b0cea1cbd947a733b563eac1b5/contracts/optimistic-ethereum/OVM/precompiles/OVM_L1MessageSender.sol#L12-L13

But I like that more granular standard approach you suggest.

contracts/optimistic-ethereum/OVM/execution/OVM_SafetyChecker.sol
```diff
@@ -6,6 +6,14 @@ import { iOVM_SafetyChecker } from "../../iOVM/execution/iOVM_SafetyChecker.sol"
...
+ * are specific to the environment (ie. L1 or L2) in which it is executed, as this could be used
+ * to "escape the sandbox" of the OVM, resulting in non-deterministic fraud proofs. 
+ * That is, an attacker wouuld be able to "prove fraud" on an honestly applied transaction.
```
ben-chain, line 13:
```suggestion
 * That is, an attacker would be able to "prove fraud" on an honestly applied transaction.
 * Note that a "safe" contract requires opcodes to appear in a particular pattern; omission of "unsafe" opcodes is necessary, but not sufficient.
```

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -24,6 +24,17 @@ import { OVM_FraudContributor } from "./OVM_FraudContributor.sol";
 
 /**
  * @title OVM_StateTransitioner
+ * @dev The OVM_StateTransitioner coordinates the execution of a state transition during the evaluation of a 
+ * fraud proof. It controls the OVM_ExecutionManager, as well as an instance of the OVM_StateManager (which is 
+ * uniquely created for each fraud proof).
+ * Once a fraud proof has been initialized, this contract is populated with the pre-state root as
+ * well as the OVM storage slots used in the transaction whose state commitment is being proven fraudulent.
```
ben-chain, line 31: To disambiguate, I would indicate that it is specifically the SM which gets populated with the OVM storage slots.  This contract authenticates/verifies that those populations check out with the `preStateRoot`, but it does not store them itself.

ben-chain (commented):
This is coming along!  Left a few suggestions.

Last meta-note on what to call these things in comments: I think "State/Execution Manager" is fine vs. specifying the exact contract name.  In the extreme, I actually think EM/SM is such nice shorthand for those critical contracts.  I would even be open to even using that inline, so long as we explicitly state somewhere in the docs.  Curious what you think -- that might be bad practice; we can't guarantee folks will read.

maurelian:
These checks are failing, because of something to do with the ovm compiler:
image

https://github.com/ethereum-optimism/contracts-v2/runs/1785015600?check_suite_focus=true

ben-chain:
About to review the diff, but off the top, I would try re-running CI and see if it just goes away.  I sometimes get timeouts when downloading the `0.7.6` compiler off of github, even locally.  Could be that + ungraceful error handling.

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
```diff
@@ -13,6 +13,12 @@ import { Lib_SafeMathWrapper } from "../../libraries/wrappers/Lib_SafeMathWrappe
 
 /**
  * @title OVM_ECDSAContractAccount
+ * @dev The ECDSA Contract Account exists for backwards compatibility with Ethereum's Layer 1.
+ * It enables the owner of a Layer 1 account to prove ownership and send transaction from that address on Layer 2.
```
ben-chain, line 17: A nit would be that the `ovmCREATEEOA` opcode is what enables sending transactions *from that address*, and this contract makes it so that the transaction format is backwards compatible with ETH1.

Could also be worth noting here that the alternative `eth_sign` scheme is also supported by this contract.

contracts/optimistic-ethereum/OVM/bridge/OVM_L1CrossDomainMessenger.sol
```diff
@@ -19,7 +19,13 @@ import { OVM_BaseCrossDomainMessenger } from "./OVM_BaseCrossDomainMessenger.sol
 
 /**
  * @title OVM_L1CrossDomainMessenger
- * @dev This contract lives on L1. It sends L1->L2 messages into L2, and relays L2->L1 messages from L2 to their target on L1.
+ * @dev The L1 Cross Domain Messenger contract sends messages from L1 to L2, and relays messages from L2 onto L1. 
+ * In the event that a message sent from L1 to L2 is rejected for exceeding the L2 epoch gas limit, it can be resubmitted 
+ * via this contracts replay function.
```
ben-chain, line 24:
```suggestion
 * via this contract's replay function. 
```

contracts/optimistic-ethereum/OVM/bridge/OVM_L2CrossDomainMessenger.sol
```diff
@@ -17,9 +17,12 @@ import { OVM_BaseCrossDomainMessenger } from "./OVM_BaseCrossDomainMessenger.sol
 
 /**
  * @title OVM_L2CrossDomainMessenger
- * @dev L2 CONTRACT (COMPILED)
- * This contract lives on L2. It sends messages to L1, and relays them from L1.
- */
+ * @dev The L2 Cross Domain Messenger contract sends messages from L2 to L1, and relays messages from L1 onto L1.
```
ben-chain, line 20:
```suggestion
 * @dev The L2 Cross Domain Messenger contract sends messages from L2 to L1, and is the entry point for L2 messages sent via the L1 Cross Domain Messenger.
```
What I'm trying to get at here is that the relaying is passive, and that `enqueue` can be used to send L1->L2 messages which bypass this contract, so it is not *always* relaying the messages.

contracts/optimistic-ethereum/OVM/bridge/OVM_L2CrossDomainMessenger.sol
```diff
@@ -17,9 +17,12 @@ import { OVM_BaseCrossDomainMessenger } from "./OVM_BaseCrossDomainMessenger.sol
...
+ * 
+ * Compiler used: optimistic-solc
+ * Runtime target: OVM
+ * Execution environment: Layer 2
```
ben-chain, line 24: I'm a little confused by the difference between these two fields.  To me they feel very analogous.

Is it worth adding a `style-guide.md` somewhere which delineates the standard and explains what the common fields in docs should mean?

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -19,6 +19,19 @@ import { OVM_DeployerWhitelist } from "../precompiles/OVM_DeployerWhitelist.sol"
...
 /**
  * @title OVM_ExecutionManager
+ * @dev The Execution Manager (EM) is the core of our OVM implementation, and provides a sandboxed 
+ * environment allowing us to execute OVM transactions deterministically on either Layer 1 or 
+ * Layer 2. 
+ * The EM's run() function is the first function called during the execution of any
+ * transaction on L2. 
+ * For each context-dependent EVM operation the EM has a function which implements a corresponding 
+ * OVM operation, which will read state from the State Manager contract.
+ * The EM relies on the Safety Checker to verify that code running on Layer 2 does not contain any
```
ben-chain, line 29:
```suggestion
 * The EM relies on the Safety Checker to verify that code deployed to Layer 2 does not contain any 
```
The reason for this distinction is that we do not safety check contracts which are already deployed in the `preState`.  (Of course, if that is the case, we should be proving fraud on the earlier block in which the unsafe contract deploy was attempted).

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -19,6 +19,19 @@ import { OVM_DeployerWhitelist } from "../precompiles/OVM_DeployerWhitelist.sol"
 
 /**
  * @title OVM_ExecutionManager
+ * @dev The Execution Manager (EM) is the core of our OVM implementation, and provides a sandboxed 
+ * environment allowing us to execute OVM transactions deterministically on either Layer 1 or 
+ * Layer 2. 
+ * The EM's run() function is the first function called during the execution of any
+ * transaction on L2. 
+ * For each context-dependent EVM operation the EM has a function which implements a corresponding 
+ * OVM operation, which will read state from the State Manager contract.
```
ben-chain, line 28: nit: some of these do not read from the SM, instead reading from EM storage (messageContext and globalContext).  Not sure the succinct way to express.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -153,6 +166,7 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         // OVM_StateManager (expected to be an OVM_StateTransitioner). We can revert here because
         // this would make the `run` itself invalid.
         require(
+            // This method always returns true in the L2 State Manager precompile
```
ben-chain, line 169:
```suggestion
            // This method enforces access during fraud proofs, but returns true in the L2 node's State Manager precompile.
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -308,8 +322,8 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
     }
 
     /**
-     * @notice Specifies what L1 EOA, if any, sent this transaction.
-     * @return _l1TxOrigin Address of the EOA which send the tx into L2 from L1.
+     * @notice Specifies which L1 account, if any, sent this transaction.
```
ben-chain, line 325:
```suggestion
     * @notice Specifies which L1 account, if any, sent this transaction by calling `enqueue`.
```

contracts/optimistic-ethereum/OVM/execution/OVM_SafetyChecker.sol
```diff
@@ -6,6 +6,17 @@ import { iOVM_SafetyChecker } from "../../iOVM/execution/iOVM_SafetyChecker.sol"
 
 /**
  * @title OVM_SafetyChecker
+ * @dev  The Safety Checker verifies that contracts deployed on L2 do not contain any
+ * "unsafe" operations. An operation is considered unsafe if it would access state variables which
+ * are specific to the environment (ie. L1 or L2) in which it is executed, as this could be used
+ * to "escape the sandbox" of the OVM, resulting in non-deterministic fraud proofs. 
+ * That is, an attacker would be able to "prove fraud" on an honestly applied transaction.
+ * Note that a "safe" contract requires opcodes to appear in a particular pattern;
```
ben-chain, line 14: "see docs for more"?

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -10,6 +10,14 @@ import { iOVM_StateManager } from "../../iOVM/execution/iOVM_StateManager.sol";
 
 /**
  * @title OVM_StateManager
+ * @dev The State Manager contract holds all storage values for contracts in the OVM. It can only be written to by the 
+ * the Execution Manager and State Transitioner. It runs on L1 during the setup and execution of a fraud proof.
```
ben-chain, line 14: Technically it is also used in the `POST_EXECUTION` phase; we read what OVM storage has been updated to generate the `postStateRoot`.

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -10,6 +10,14 @@ import { iOVM_StateManager } from "../../iOVM/execution/iOVM_StateManager.sol";
...
+ * The same logic runs on L2, but has been implemented as a precompile in the L2 go-ethereum client 
+ * (see https://github.com/ethereum-optimism/go-ethereum/blob/master/core/vm/ovm_state_manager.go)
+ * 
+ * Compiler used: solc
+ * Runtime target: EVM
+ * Execution environment: Layer 1 and Layer 2
```
ben-chain, line 20: Because this is a precompile, this contract (as in the actual bytecode) is not executed on L2 nodes, so depending on our definition of EE here, might wanna say only L1.

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -61,10 +69,11 @@ contract OVM_StateManager is iOVM_StateManager {
      **********************/
 
     /**
-     * Simple authentication, this contract should only be accessible to the owner or to the
-     * OVM_ExecutionManager during the transaction execution process.
+     * Simple authentication, this contract should only be accessible to the owner (which is expected to be the state
```
ben-chain, line 72:
```suggestion
     * Simple authentication, this contract should only be accessible to the owner (which is expected to be the State Transitioner during `PRE_EXECUTION` 
```

contracts/optimistic-ethereum/OVM/execution/OVM_StateManager.sol
```diff
@@ -61,10 +69,11 @@ contract OVM_StateManager is iOVM_StateManager {
      **********************/
 
     /**
-     * Simple authentication, this contract should only be accessible to the owner or to the
-     * OVM_ExecutionManager during the transaction execution process.
+     * Simple authentication, this contract should only be accessible to the owner (which is expected to be the state 
+     * or to theOVM_ExecutionManager during the transaction execution process.
```
ben-chain, line 73:
```suggestion
     * or the OVM_ExecutionManager during transaction execution.
```

contracts/optimistic-ethereum/OVM/execution/OVM_StateManagerFactory.sol
```diff
@@ -10,6 +10,12 @@ import { OVM_StateManager } from "./OVM_StateManager.sol";
 
 /**
  * @title OVM_StateManagerFactory
+ * @dev The State Manager Factory is called by a State Transitioner's init code, to create a new
```
ben-chain, line 13:
```suggestion
 * @dev The State Manager Factory is called by a State Transitioner's fraud proof initialization code, to create a new 
```
(don't wanna confuse with creationCode which often is called initcode)

contracts/optimistic-ethereum/OVM/precompiles/OVM_ETH.sol
```diff
@@ -11,7 +11,12 @@ import { iOVM_BaseCrossDomainMessenger } from "../../iOVM/bridge/iOVM_BaseCrossD
 
 /**
  * @title OVM_ETH
- * @dev L2 CONTRACT (COMPILED)
+ * @dev The ETH predeploy provides an ERC20 interface for ETH deposited to Layer 2. Note that 
+ * unlike on Layer 1, Layer 2 accounts do not have a balance field.
```
ben-chain, line 15: nit: we do technically have a balance field, just left it untouched in geth.  However, there is no way for it to have a nonzero value.

contracts/optimistic-ethereum/OVM/precompiles/OVM_L1MessageSender.sol
```diff
@@ -7,7 +7,19 @@ import { iOVM_ExecutionManager } from "../../iOVM/execution/iOVM_ExecutionManage
 
 /**
  * @title OVM_L1MessageSender
- * @dev L2 CONTRACT (NOT COMPILED)
+ * @dev The L1MessageSender is a predeploy contract running on L2. During the execution of cross 
+ * domain transaction from L1 to L2, it returns the address of the L1 account (either an EOA or
+ * contract) which sent the message to L2 via the Canonical Transaction Chain's `enqueue()` 
+ * function.
+ * 
+ * This contract can be thought of as a getter for the ovmL1TXORIGIN operation. This is necessary
```
ben-chain, line 15:
```suggestion
 * This contract exclusively serves as a getter for the ovmL1TXORIGIN operation. This is necessary 
```

contracts/optimistic-ethereum/OVM/precompiles/OVM_ProxySequencerEntrypoint.sol
```diff
@@ -4,7 +4,13 @@ pragma solidity ^0.7.0;
 import { Lib_SafeExecutionManagerWrapper } from "../../libraries/wrappers/Lib_SafeExecutionManagerWrapper.sol";
 
 /**
- * @title OVM_ProxySequencerEntrypoint
+ * @title OVM_ProxySequencerEntrypoint 
+ * @dev The Proxy Sequencer Entrypoint is a predeployed proxy to the implementation of the 
+ * Sequencer Entrypoint. This will enable the Optimism team to upgrade the Sequencer Entrypoint 
+ * contract.
+ * 
+ * This contract compiles to __
```
ben-chain, line 12: missing update here

contracts/optimistic-ethereum/OVM/precompiles/OVM_ProxySequencerEntrypoint.sol
```diff
@@ -4,7 +4,13 @@ pragma solidity ^0.7.0;
 import { Lib_SafeExecutionManagerWrapper } from "../../libraries/wrappers/Lib_SafeExecutionManagerWrapper.sol";
 
 /**
- * @title OVM_ProxySequencerEntrypoint
+ * @title OVM_ProxySequencerEntrypoint 
+ * @dev The Proxy Sequencer Entrypoint is a predeployed proxy to the implementation of the 
+ * Sequencer Entrypoint. This will enable the Optimism team to upgrade the Sequencer Entrypoint 
+ * contract.
```
ben-chain, line 10: could be worth noting here that this is not an admin key, since it has no safety/authentication logic -- worse we could do is change it to something that doesn't correctly supply inputs to the contract wallets.

ben-chain (commented):
This is amazing!!!  I have to jump off for 30 minutes, posting this batch of feedback until I can get to the rest.

contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
```diff
@@ -8,6 +8,15 @@ import { Lib_SafeExecutionManagerWrapper } from "../../libraries/wrappers/Lib_Sa
 
 /**
  * @title OVM_SequencerEntrypoint
+ * @dev The Sequencer Entrypoint is a predeploy which, despite its name, can in fact be called by 
+ * any account. It accepts a more efficient compressed calldata format, which it decompresses and
```
ben-chain, line 12: Heh, good catch!  I wonder if there's any problems with anyone being able to call this; I don't think so.  Nobody should want to, since it would mean the fee sent by the ECDSAAccount gets sent to this address instead of their address.

contracts/optimistic-ethereum/OVM/verification/OVM_StateTransitioner.sol
```diff
@@ -24,6 +24,18 @@ import { OVM_FraudContributor } from "./OVM_FraudContributor.sol";
 
 /**
  * @title OVM_StateTransitioner
+ * @dev The State Transitioner coordinates the execution of a state transition during the evaluation of a 
+ * fraud proof. It controls the Execution Manager, as well as an instance of the State Manager (which is
```
ben-chain, line 28:
```suggestion
 * fraud proof. It feeds verified input to the Execution Manager's run(), and controls a State Manager (which is 
```

contracts/optimistic-ethereum/libraries/wrappers/Lib_SafeExecutionManagerWrapper.sol
```diff
@@ -42,7 +42,7 @@ library Lib_SafeExecutionManagerWrapper {
     }
 
     /**
-     * Makes an ovmCALL and performs all the necessary safety checks.
+     * Makes an ovmDELEGATECALL and performs all the necessary safety checks.
```
ben-chain, line 45: Seeing this makes me remember how I really don't like the phrase we use throughout this, "`performs all the necessary safety checks`".

I think this is a little confusing because the `OVM_SafetyChecker` is what really "checks safety".  What these functions are really doing is *implementing the equivalent logic to what the safety checker looks for*.

Maybe an alternative way to say it is something like "and takes the necessary precautions to guarantee the sandbox can't be escaped", with a little more detail explaining the above in the header.

ben-chain (commented):
These are all super amazing!!  Left a round of feedback, very excited for this to get in.

contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
maurelian: No, I don't think so. This actually came up in a conversation we had... 

But I'm going to consider that out of scope. :)

contracts/optimistic-ethereum/libraries/wrappers/Lib_SafeExecutionManagerWrapper.sol
maurelian: I addressed this by adding a descriptive header comment: 
```
/**
 * @title Lib_SafeExecutionManagerWrapper
 * @dev The Safe Execution Manager Wrapper provides functions which facilitate writing OVM safe 
 * code using the standard solidity compiler, by routing all its operations through the Execution 
 * Manager.
 * 
 * Compiler used: solc
 * Runtime target: OVM
 */
 ```
 
 Then just making the function comments verify brief, ie. `Performs a safe ovmCALL.`.

**Comment by maurelian** [comment]
OMG, the CI is happy, ship it!

File: contracts/optimistic-ethereum/OVM/precompiles/OVM_ProxySequencerEntrypoint.sol [hunk 27]
```diff
@@ -5,7 +5,13 @@ pragma solidity >0.5.0 <0.8.0;
...
+ * 
+ * This contract compiles to __
+ * It is (only) deployed on Layer ...
```
ben-chain, line 14:
```suggestion
 * Compiler used: solc
 * Runtime target: OVM
```

ben-chain (approved):
This is awesome, thanks again!!

PR #205, opened 2021-01-27
ctc: less calls

tynes:
## Description

Less calls for `appendSequencerBatch`. This should reduce the number of calls by about half.

There is also some low hanging fruit in `appendQueueBatch` but was left for another PR since that functionality is currently disabled. 

## Contributing Agreement


maurelian (approved):
This change simply caches the address returned by `queue()` with a given function call, thus saving making two calls to the Address Resolver. 

LGTM.

PR #207, opened 2021-02-01
feat/multirelay

tynes:
## Description

Adds a `MultiRelay` contract that can be set as `OVM_L2MessageRelayer` in the address manager that can call `relayMessage()` many times. This will greatly reduce the number of transactions that need to be sent when relaying L2 to L1 messages.

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
@@ -55,8 +55,7 @@ contract OVM_L1MultiMessageRelayer is iOVM_L1MultiMessageRelayer, Lib_AddressRes
         external
         onlyBatchRelayer 
     {
-        // @todo: confirm correct string (this or proxy?)
-        iOVM_L1CrossDomainMessenger messenger = iOVM_L1CrossDomainMessenger(resolve("OVM_L1CrossDomainMessenger"));
+        iOVM_L1CrossDomainMessenger messenger = iOVM_L1CrossDomainMessenger(resolve("Proxy__OVM_L1CrossDomainMessenger"));
         for (uint256 i = 0; i < _messages.length; i++) {
             L2ToL1Message memory message = _messages[i];
             messenger.relayMessage(
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
@@ -18,7 +18,6 @@ import { Lib_AddressResolver } from "../../libraries/resolver/Lib_AddressResolve
  *
  * Compiler used: solc
  * Runtime target: EVM
- * Execution environment: Layer 1
  */
 contract OVM_L1MultiMessageRelayer is iOVM_L1MultiMessageRelayer, Lib_AddressResolver {
 
--- a/contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+// @unsupported: ovm
+pragma solidity >0.5.0 <0.8.0;
+pragma experimental ABIEncoderV2;
+/* Interface Imports */
+import { iOVM_L1CrossDomainMessenger } from "../../iOVM/bridge/iOVM_L1CrossDomainMessenger.sol";
+import { iOVM_L1MultiMessageRelayer } from "../../iOVM/bridge/iOVM_L1MultiMessageRelayer.sol";
+
+/* Contract Imports */
+import { Lib_AddressResolver } from "../../libraries/resolver/Lib_AddressResolver.sol";
+
+
+/**
+ * @title OVM_L1MultiMessageRelayer
+ * @dev The L1 Multi-Message Relayer contract is a gas efficiency optimization which enables the 
+ * relayer to submit multiple messages in a single transaction to be relayed by the L1 Cross Domain
+ * Message Sender.
+ *
+ * Compiler used: solc
+ * Runtime target: EVM
+ * Execution environment: Layer 1
+ */
+contract OVM_L1MultiMessageRelayer is iOVM_L1MultiMessageRelayer, Lib_AddressResolver {
+
+    /***************
+     * Constructor *
+     ***************/
+    constructor(
+        address _libAddressManager
+    ) 
+        Lib_AddressResolver(_libAddressManager)
+    {}
+
+    /**********************
+     * Function Modifiers *
+     **********************/
+
+    modifier onlyBatchRelayer() {
+        require(
+            msg.sender == resolve("OVM_L2BatchMessageRelayer"),
+            "OVM_L1MultiMessageRelayer: Function can only be called by the OVM_L2BatchMessageRelayer"
+        );
+        _;
+    }
+
+    /********************
+     * Public Functions *
+     ********************/
+
+    /**
+     * @notice Forwards multiple cross domain messages to the L1 Cross Domain Messenger for relaying
+     * @param _messages An array of L2 to L1 messages
+     */
+    function batchRelayMessages(L2ToL1Message[] calldata _messages) 
+        override
+        external
+        onlyBatchRelayer 
+    {
+        // @todo: confirm correct string (this or proxy?)
+        iOVM_L1CrossDomainMessenger messenger = iOVM_L1CrossDomainMessenger(resolve("OVM_L1CrossDomainMessenger"));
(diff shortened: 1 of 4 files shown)


contracts/optimistic-ethereum/OVM/bridge/OVM_MultiRelay.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    struct L2ToL1Message {
+        address target;
+        address sender;
+        bytes message;
+        uint256 messageNonce;
+        iOVM_L1CrossDomainMessenger.L2MessageInclusionProof proof;
+    }
+
+    address public relay;
```
tynes, line 27: I think this could be a `iOVM_L1CrossDomainMessenger` instead of just an `address`

src/contract-deployment/config.ts
```diff
@@ -110,6 +110,10 @@ export const makeContractDeployConfig = async (
         )
       },
     },
+    OVM_MultiRelay: {
+      factory: getContractFactory('OVM_MultiRelay'),
+      params: [await AddressManager.getAddress('Proxy__OVM_L1CrossDomainMessenger')]
```
tynes, line 115: This results in a race condition when not waiting for the transactions and could initialize the relay to the zero address

src/contract-deployment/config.ts
```diff
@@ -110,6 +110,12 @@ export const makeContractDeployConfig = async (
         )
       },
     },
+    OVM_MultiRelay: {
+      factory: getContractFactory('OVM_MultiRelay'),
+      params: [
+        await AddressManager.getAddress('Proxy__OVM_L1CrossDomainMessenger'),
```
maurelian, line 116: Does the double underscore have some significance?

contracts/optimistic-ethereum/OVM/bridge/OVM_MultiRelay.sol
maurelian: I agree that'd be better.

ben-chain (commented):
Is might be worth resolving the owner here from the Address Manager--could make future "changes of the guard" a little easier.

maurelian (commented):
Would it be difficult to add a test to this? That'd at least facilitate some future gas regression tracking.
I could take that on with a pointer in the right direction.

tynes:
> Is might be worth resolving the owner here from the Address Manager--could make future "changes of the guard" a little easier.

I agree with this because everything else in the system is using the Address Manager

> Would it be difficult to add a test to this? 

Agree with this as well, adding this trades some gas for developer time. No tests so its just a draft PR still :)

maurelian:
Per the stand up this morning, I'll take this task over.

contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @notice Forwards multiple cross domain messages to the L1 Cross Domain Messenger for relaying
+     * @param _messages An array of L2 to L1 messages
+     */
+    function batchRelayMessages(L2ToL1Message[] calldata _messages) 
+        override
+        external
+        onlyBatchRelayer 
+    {
+        // @todo: confirm correct string (this or proxy?)
```
ben-chain, line 58: Hmmmm... it actually looks like there currently is no entry for the proxy itself.  It looks like we set the implementation as the `OVM_L1CrossDomainMessenger` (https://github.com/ethereum-optimism/contracts-v2/blob/bb91c16d0c4de513a023fb6d4512bfb1a6c85999/src/contract-deployment/config.ts#L84) and then the proxy uses that but is actually not put into the AM.  So, we will need to set a new address for the proxy, as well as get the new address here.

test/contracts/OVM/bridge/OVM_L1MultiMessageRelayer.ts
```diff
@@ -0,0 +1,134 @@
...
+    } 
+
+    // create a few dummy messages to relay
+    let m1 = {
+      target: "0x1100000000000000000000000000000000000000",
+      message: NON_NULL_BYTES32,
+      sender: "0x2200000000000000000000000000000000000000",
+      messageNonce: 1,
+      proof: dummyProof,
+      calldata: "0xaabbccddeeff"
```
ben-chain, line 70: I think these fields are not needed -- I don't see any `calldata` field in the contract structs.  Note that the `message` should contain the data being sent around.

test/contracts/OVM/bridge/OVM_L1MultiMessageRelayer.ts
```diff
@@ -0,0 +1,134 @@
...
+  })
+
+  describe('batchRelayMessages', () => {
+    it('Successfully relay multiple messages', async () => {
+      await OVM_L1MultiMessageRelayer.batchRelayMessages(
+        messages
+      )
+      await expect(
+        Mock__OVM_L1CrossDomainMessenger.smocked.relayMessage.calls.length
+      ).to.deep.equal(messages.length)
```
ben-chain, line 116: Nit: I think we should be able to explicitly assert `(calls).to.deep.equal(messages)` here if the types work out how I'm thinking.

ben-chain (commented):
Lookin' good!  Left a couple small notes.  I think CI will be hard to get passing without updating some of the off-chain services which use this.  Maybe we can just pass that off to @kfichter when the time comes 😇

test/contracts/OVM/bridge/OVM_L1MultiMessageRelayer.ts
maurelian: It doesn't seem to be straightforward. I tried `deep.equal` and `deep.include` but neither are working. I also tried iterating  over `calls` and `messages` with no luck. 

Each message looks like this:

```
{ target: '0x1100000000000000000000000000000000000000',
    message: '0x1111111111111111111111111111111111111111111111111111111111111111',
    sender: '0x2200000000000000000000000000000000000000',
    messageNonce: 1,
    proof:
     { stateRoot: '0x1111111111111111111111111111111111111111111111111111111111111111',
       stateRootBatchHeader: [Object],
       stateRootProof: [Object],
       stateTrieWitness: '0x736f6d65206279746573',
       storageTrieWitness: '0x736f6d65206d6f7265206279746573' } }
```

Whereas the calls object looks like the attached, with mangled keys for the arguments: 
[calls.txt](https://github.com/ethereum-optimism/contracts-v2/files/5914585/calls.txt)

Can it be done, yes, and it'd be nice to have built into smock to decode the calls object, but it's not worth the effort right now IMO.

tynes:
Did you change something about how the contracts compile? From the integration tests:

```
cat: /opt/contracts-v2/cache/last-solc-config.json: No such file or directory
```

The deployer docker image has a little automation around sending a few files to a special hardhat endpoint to register those contracts so that calls can log the name of the contract being called + the method. This ideally should happen in the JS so that its not coupled to the docker stuff, see https://github.com/ethereum-optimism/docker/blob/d6149ffe7e8453f0cf7352b5b9ac8b913fa4229f/deployer/wait-for-l1.sh#L27

maurelian:
@tynes I think that might be a change from #208, which moved away from using `// +build: ovm` on contracts to opt-in to the ovm compiler, and moved towards using `// @unsupported: ovm` on contracts to opt-out of ovm-compilation.

contracts/optimistic-ethereum/OVM/bridge/OVM_L1MultiMessageRelayer.sol
kfichter: Proxy entry is `Proxy__OVM_L1CrossDomainMessenger`

tynes:
> @tynes I think that might be a change from #208, which moved away from using `// +build: ovm` on contracts to opt-in to the ovm compiler, and moved towards using `// @unsupported: ovm` on contracts to opt-out of ovm-compilation.

Will need to update the docker repo to use the new locations of the files

maurelian:
I don't think the change to `// @unsupported: ovm` would affect the CTC. Maybe I broke something in the deployment config.

tynes:
This is working on mainnet btw

tynes:
Not sure why the tests are breaking here, but this needs to be merged ASAP so it doesn't fall behind and become stale as it is deployed to mainnet

test/contracts/OVM/bridge/OVM_L1MultiMessageRelayer.ts
```diff
@@ -0,0 +1,126 @@
...
+    await AddressManager.setAddress(
+      'OVM_L2MessageRelayer', // This is the string currently used in the AddressManager
+      OVM_L1MultiMessageRelayer.address
+    )
+    // set the mock return value
+    Mock__OVM_L1CrossDomainMessenger.smocked.relayMessage.will.return()
+  })
+
+  describe('batchRelayMessages', () => {
+    it('Successfully relay multiple messages', async () => {
```
K-Ho, line 105: @maurelian looks like this tests is failing

ben-chain (approved):
Thanks for cleaning up the CI, let's ship it!

PR #210, opened 2021-02-02
Various optimizations to appendSequencerBatch

kfichter:
## Description
Improves `appendSequencerBatch` so the gas overhead per transaction (beyond inherent `calldata` cost) is on the order of 1-2k gas.

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -16,6 +16,8 @@ import { iOVM_ChainStorageContainer } from "../../iOVM/chain/iOVM_ChainStorageCo
 /* Contract Imports */
 import { OVM_ExecutionManager } from "../execution/OVM_ExecutionManager.sol";
 
+import { console } from "hardhat/console.sol";
+
 
 /**
  * @title OVM_CanonicalTransactionChain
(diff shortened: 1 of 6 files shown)


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -384,6 +374,7 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         );
 
         // Get queue length for future comparison/
+        iOVM_ChainStorageContainer queueRef = queue();
```
tynes, line 377: Consider these changes to cut down on calls to the address manager: https://github.com/ethereum-optimism/contracts-v2/pull/205/files#diff-c3f43bf36b5d31fd82bda152c3e2fd95880d2ded92a46c7363b8463d6d2acd0fR161-R164

`getQueueLength` has a `resolve()` in it, i think we should never call it internally

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -16,6 +16,8 @@ import { iOVM_ChainStorageContainer } from "../../iOVM/chain/iOVM_ChainStorageCo
 /* Contract Imports */
 import { OVM_ExecutionManager } from "../execution/OVM_ExecutionManager.sol";
 
+import { console } from "hardhat/console.sol";
```
ben-chain, line 19: whoopsie

PR #211, opened 2021-02-05
Improving comments and some names

maurelian:
## Description

The main changes are:

- Renamed two abstract contracts (and file names) from `OVM_*` to `Abs_`
- copying docstrings into the CTC impl
- Added a missing function to the CTC interface, but that broke the ovm-compiler, so I left it there but commented it out.


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -189,7 +194,8 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
     }
 
     /**
-     * @inheritdoc iOVM_CanonicalTransactionChain
+     * Get the number of queue elements which have not yet been included.
+     * @return Length of the queue.
```
ben-chain, line 198: I think this is left over from a copy-paste of below?

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -220,7 +228,10 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
     }
 
     /**
-     * @inheritdoc iOVM_CanonicalTransactionChain
+     * Adds a transaction to the queue.
+     * @param _target Target contract to send the transaction to.
+     * @param _gasLimit Gas limit for the given transaction.
```
ben-chain, line 233:
```suggestion
     * @param _target Target L2 contract to send the transaction to.
     * @param _gasLimit Gas limit for the enqueued L2 transaction.
```

contracts/optimistic-ethereum/iOVM/chain/iOVM_CanonicalTransactionChain.sol
```diff
@@ -60,6 +63,19 @@ interface iOVM_CanonicalTransactionChain {
...
+     * @return Reference to the queue storage container.
+     */
+     // @todo: uncommenting this line breaks `yarn build`, but not `solc`. 
+     // must be something in the ovm-compiler, but why?
+    // function queue()
+    //     external
+    //     view
+    //     returns (
+    //         iOVM_ChainStorageContainer
+    //     );
```
ben-chain, line 77: My best guess: the `ovm-compiler` plugin doesn't handle errors as gracefully, but under the hood, it is complaining that `queue()` is not marked `override` in the implementation.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
maurelian: It was ported directly from the interface comments, good catch.

contracts/optimistic-ethereum/iOVM/chain/iOVM_CanonicalTransactionChain.sol
maurelian: Correct, that resolved it.

kfichter:
This looks good to me. I'll give an implicit approval but will wait for @ben-chain to actually hit the "approve" button.

ben-chain (approved):
Great, LGTM now! 🚀

PR #218, opened 2021-02-10
Fix package json so the build actually works

kfichter:
## Description
We moved over to `hardhat`, which means that the structure of the `artifacts` folder has changed. As a result, our package.json is broken and our artifacts are not being included as part of the `build` folder published to `npm`. We need this update or we can't use any artifacts from our latest contract versions.

## Contributing Agreement


karlfloersch (approved):
LGTM!

PR #219, opened 2021-02-10
Standard Deposit/Withdrawal Interface

ben-chain:
## Description
This PR adds a standard interface for the minimal pair of contracts required for deposit/withdrawals.  If there's any wishlist functionality that could go in here, we can discuss, but I lean towards letting developers expand this out as needed if they desire bespoke functionality that would impose potential security risks.  For example, considered some sort of `depositAndCall`, but I think the security implications of this mean we should just have people add it to theirs as needed.

## Questions
- Anything else we might want here?

### Fixes
- Fixes roadmap 557

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L1ERC20Gateway.sol
+++ b/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L1ERC20Gateway.sol
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >0.5.0 <0.8.0;
+pragma experimental ABIEncoderV2;
+/**
+ * @title iOVM_L1ERC20Gateway
+ */
+interface iOVM_L1ERC20Gateway {
+
+    /**********
+     * Events *
+     **********/
+    event DepositInitiated(address indexed _from, address _to, uint256 _amount);    
+    event WithdrawalFinalized(address indexed _to, uint256 _amount);
+
+
+    /********************
+     * Public Functions *
+     ********************/
+    function deposit(uint _amount) external;
+    function depositTo(address _to, uint _amount) external;
+
+    /*************************
+     * Cross-chain Functions *
+     *************************/
+    function finalizeWithdrawal(address _to, uint _amount) external;
+}
--- a/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L1ERC20Gateway.sol
+++ b/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L1ERC20Gateway.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >0.5.0 <0.8.0;
 pragma experimental ABIEncoderV2;
+
 /**
  * @title iOVM_L1ERC20Gateway
  */
@@ -11,7 +12,6 @@ interface iOVM_L1ERC20Gateway {
      **********/
     event DepositInitiated(address indexed _from, address _to, uint256 _amount);    
     event WithdrawalFinalized(address indexed _to, uint256 _amount);
-
 
     /********************
      * Public Functions *
--- a/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L2ERC20Gateway.sol
+++ b/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L2ERC20Gateway.sol
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >0.5.0 <0.8.0;
+pragma experimental ABIEncoderV2;
+
+import { iOVM_ERC20 } from "../../precompiles/iOVM_ERC20.sol";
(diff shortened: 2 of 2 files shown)


K-Ho (commented):
looks good! Synthetix was using `completeWithdrawal` and `completeDeposit` instead of `finalize`, and I kind of like complete more? Not super opinionated here though

contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L2ERC20Gateway.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >0.5.0 <0.8.0;
+pragma experimental ABIEncoderV2;
+
+import { iOVM_ERC20 } from "../../precompiles/iOVM_ERC20.sol";
+
```
tynes, line 6: Nit: extra spacing here when there is no extra spacing in the other contract directly below the imports. Consistency makes things easier to read/audit

contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L1ERC20Gateway.sol
```diff
@@ -0,0 +1,26 @@
...
+ * @title iOVM_L1ERC20Gateway
+ */
+interface iOVM_L1ERC20Gateway {
+
+    /**********
+     * Events *
+     **********/
+    event DepositInitiated(address indexed _from, address _to, uint256 _amount);    
+    event WithdrawalFinalized(address indexed _to, uint256 _amount);
+
```
tynes, line 14: NIt: two spaces here where each other has 1

contracts/optimistic-ethereum/iOVM/bridge/ERC20/iOVM_L2ERC20Gateway.sol
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @title iOVM_L2ERC20Gateway
+ */
+interface iOVM_L2ERC20Gateway  /* TODO: add `is iOVM_ERC20` -- the types are messing with my tempERC20 */ {
```
tynes, line 11: Nit: this todo makes the line too long, could easily go in the above comment below `@title`

TransmissionsDev:
I think a `depositAndCall` function would be a great add. I don't see why risk is a concern there because devs are opting in to using that (vs like an ERC777 that just does the call no matter what). 

Other thing is an "approveDepositTo" "approveWithdrawTo" pair

ben-chain:
> Other thing is an `approveDepositTo` `approveWithdrawTo` pair of funcs (esp if you're not going to add a `depositAndCall` func)

Ooh this is very interesting! @maurelian and I had brainstormed `approveDepositBy`; ultimately this didn't make sense because an `approve`d L1 contract could just `transferFrom` and `depositTo`.  But this sounds like approving depositing to a specific target, which is very interesting. :)

Can you walk me through a use case or sketch for the underlying functionality?  Is this kind of like saying "this `1address` can deposit on my behalf, IFF it is to this `l2address`?"

I guess the concern with `depositAndCall` was that this would lead to scenarios where the L2 ERC20 can be a `msg.sender`, which is a little weird.  OTOH I guess this would normally just deliver a `data` payload wrapped in some methodId like `OVM_onDeposit(depositer,data)` implemented by the target.  At that point, it feels hard to exploit unless there's a collision in the methodID or something crazy in a fallback.

TransmissionsDev:
> Can you walk me through a use case or sketch for the underlying functionality? Is this kind of like saying "this 1address can deposit on my behalf, IFF it is to this l2address?"

That's interesting but I was just imagining those funcs having the same general implementation as `withdrawTo`/`depositTo` but just approving the token to the recipient instead of transferring it to them. Reason I think this is an important add is because if the recipient is a contract on l2 not a user, that contract will not know who set the tokens (hence the need for the approve() and transferFrom() flow we've all come to know and put up with 😅)

TransmissionsDev:
> I guess the concern with depositAndCall was that this would lead to scenarios where the L2 ERC20 can be a msg.sender, which is a little weird. OTOH I guess this would normally just deliver a data payload wrapped in some methodId like OVM_onDeposit(depositer,data) implemented by the target. At that point, it feels hard to exploit unless there's a collision in the methodID or something crazy in a fallback.

Not really sure what you mean here, maybe we can touch on this in the call as well.

ben-chain:
@TransmissionsDev What would you think is the right way to handle the if the call part of `withdrawAndCall` fails?  Seems like you either only credit the withdrawal if the subcall succeeds, or you just always credit the withdrawal.  Honestly both seem like they would have associated dangers.

TransmissionsDev:
Hmm I'm not really opinionated either way right now? I think a reversion makes the most sense and will be easiest to discover on chain, and tokens can always be force given to the target via the default transfers?

ben-chain:
This was merged in under another PR.  Most recent update to this interface was here: https://github.com/ethereum-optimism/contracts/pull/297

PR #225, opened 2021-02-13
Add ETH deposit gateway (OVM_L1ETHGateway)

gigamesh:
## Questions
- 
-
-

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iWETH.sol
+++ b/contracts/optimistic-ethereum/iOVM/bridge/ERC20/iWETH.sol
@@ -1,0 +1,9 @@
+pragma solidity >=0.5.0;
+
+interface iWETH {
+    function deposit() external payable;
+
+    function transfer(address to, uint256 value) external returns (bool);
+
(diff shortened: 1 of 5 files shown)


contracts/optimistic-ethereum/iOVM/bridge/ERC20/iWETH.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity >=0.5.0;
+
+interface iWETH {
+    function deposit() external payable;
+
+    function transfer(address to, uint256 value) external returns (bool);
+
+    function withdraw(uint256) external;
+}
```
ben-chain, line 9: Is this vestigial?  Don't see it used elsewhere.

ben-chain (approved):
Other than an interface which I assume doesn't need to be there any more, looks good to me!  (I'll have to re-approve after that new commit)  Seems like there should be a way to combine the ERC20 and ETH tests... I'll think about it a little bit before merging the big gateway PR.

PR #227, opened 2021-02-15
perf: example gas cost optimization by caching addrs

gakonst:
As title. The proper way to do this would be to refactor all of these state fetching functions to a Library with pure functions. This would give us nice separation where we can do any sloads/calls in explicit parts of the main Contract, and do all the data mangling inside the Library in memory.


ben-chain:
@gakonst We are about to merge in a big optimization here: https://github.com/ethereum-optimism/contracts/pull/228

Consensus was to avoid this sort of caching logic in favor of keeping the upgrade process maximally simple.  LMK if you think there's optimizations missing in the above or something else you want to raise with this PR and I can re-open.

PR #228, opened 2021-02-16
feat: CanonicalTransactionChain gas optimizations

smartcontracts:
## Description
Makes the following optimizations to the `CanonicalTransactionChain`:
- A single reference to `queue` is now used instead of repeated resolutions of the value.
- Similarly, a single reference to `queueLength` is used, since the length does not change during execution.
- A chunk of memory is reserved in advance to avoid the gas overhead of allocating a lot of memory for hashes.

I also went ahead and made a few minor cleanups while I was at it.

## Contributing Agreement


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -177,24 +177,10 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
-            elementTimestamp   :=         and(timestampAndBlockNumber, 0x000000000000000000000000000000000000000000000000000000FFFFFFFFFF)
-            elementBlockNumber := shr(40, and(timestampAndBlockNumber, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF0000000000))
-        }
-
-        return Lib_OVMCodec.QueueElement({
-            queueRoot: queueRoot,
-            timestamp: elementTimestamp,
-            blockNumber: elementBlockNumber
-        });
+        return _getQueueElement(
```
kfichter, line 180: Use this internal function to avoid duplicate logic.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -295,14 +280,14 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
             timestampAndBlockNumber := or(timestampAndBlockNumber, shl(40, number()))
         }
 
-        iOVM_ChainStorageContainer queue = queue();
+        iOVM_ChainStorageContainer queueRef = queue();
```
kfichter, line 283: Renamed to `queueRef` as to avoid shadowing the `queue` function.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -323,45 +308,45 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
-            lastElement.timestamp,
-            lastElement.blockNumber
-        );
-
-        emit QueueBatchAppended(
-            nextQueueIndex - _numQueuedTransactions,
-            _numQueuedTransactions,
-            getTotalElements()
-        );
+        // _numQueuedTransactions = Lib_Math.min(_numQueuedTransactions, getNumPendingQueueElements());
```
kfichter, line 314: Comment this out to make more clear that it is unused.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -406,54 +391,83 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
 
         uint40 nextTransactionPtr = uint40(BATCH_CONTEXT_START_POS + BATCH_CONTEXT_SIZE * numContexts);
-        uint256 calldataSize;
-        assembly {
-            calldataSize := calldatasize()
-        }
 
         require(
-            calldataSize >= nextTransactionPtr,
+            msg.data.length >= nextTransactionPtr,
```
kfichter, line 396: `msg.data.length == calldatasize` and it allows me to avoid "stack too deep".

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -474,24 +488,28 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
 
         // Generate the required metadata that we need to append this batch
         uint40 numQueuedTransactions = totalElementsToAppend - numSequencerTransactions;
-        uint40 timestamp;
+        uint40 blockTimestamp;
```
kfichter, line 491: Rename to avoid shadowing `timestamp` builtin.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -674,11 +692,47 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
+        internal
+        view
+        returns (
+            Lib_OVMCodec.QueueElement memory _element
+        )
+    {
+        (
+            bytes32 queueRoot,
+            bytes32 timestampAndBlockNumber
+        ) = _queueRef.get2(uint40(_index * 2));
```
kfichter, line 713: Make use of `get2` to reduce number of `STATICCALL`s.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -701,22 +755,25 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
+        bytes memory _hashMemory
     )
         internal
         pure
         returns (
             bytes32
         )
     {
+        if (BYTES_TILL_TX_DATA + _txDataLength > _hashMemory.length) {
+            _hashMemory = new bytes(BYTES_TILL_TX_DATA + _txDataLength);
```
kfichter, line 768: Only allocate more memory if we didn't reserve enough to begin with.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -794,10 +851,11 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
     )
         internal
     {
-        (uint40 totalElements, uint40 nextQueueIndex, uint40 lastTimestamp, uint40 lastBlockNumber) = _getBatchExtraData();
+        iOVM_ChainStorageContainer batchesRef = batches();
+        (uint40 totalElements, uint40 nextQueueIndex,,) = _getBatchExtraData();
```
kfichter, line 855: Remove unused variables from function output.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -836,12 +894,28 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         // If there are existing elements, this batch must come later.
         if (getTotalElements() > 0) {
             (,, uint40 lastTimestamp, uint40 lastBlockNumber) = _getBatchExtraData();
-            require(_firstContext.blockNumber >= lastBlockNumber, "Context block number is lower than last submitted.");
-            require(_firstContext.timestamp >= lastTimestamp, "Context timestamp is lower than last submitted.");
+    
+            require(
```
kfichter, line 898: Very minor cleanup work.

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.gas.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+        ],
+        transactions,
+      })
+
+      const receipt = await res.wait()
+
+      console.log('Benchmark complete.')
+      console.log('Gas used:', receipt.gasUsed.toNumber())
+      console.log('Fixed calldata cost:', fixedCalldataCost)
+      console.log(
```
kfichter, line 286: Output **after** this PR:

```
Benchmark: 200 transactions in a single context.
Benchmark complete.
Gas used: 1641438
Fixed calldata cost: 1280000
Non-calldata overhead gas cost per transaction: 1807.19
```

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.gas.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+        }),
+        transactions,
+      })
+
+      const receipt = await res.wait()
+
+      console.log('Benchmark complete.')
+      console.log('Gas used:', receipt.gasUsed.toNumber())
+      console.log('Fixed calldata cost:', fixedCalldataCost)
+      console.log(
```
kfichter, line 326: Output **after** this PR:

```
Benchmark: 200 transactions in 200 contexts.
Benchmark complete.
Gas used: 1813692
Fixed calldata cost: 1280000
Non-calldata overhead gas cost per transaction: 2668.46
```

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.gas.spec.ts
kfichter: Output **before** this PR:

```
Benchmark: 200 transactions in a single context.
Benchmark complete.
Gas used: 1721736
Fixed calldata cost: 1280000
Non-calldata overhead gas cost per transaction: 2208.68
```

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.gas.spec.ts
kfichter: Output **before** this PR:

```
Benchmark: 200 transactions in 200 contexts.
Benchmark complete.
Gas used: 3709053
Fixed calldata cost: 1280000
Non-calldata overhead gas cost per transaction: 12145.265
```

kfichter (commented):
Added comments for clarity

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.gas.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+}
+
+interface BatchContext {
+  numSequencedTransactions: number
+  numSubsequentQueueTransactions: number
+  timestamp: number
+  blockNumber: number
+}
+
+interface AppendSequencerBatchParams {
```
tynes, line 81: We should consider combining this stuff into `core-utils` so that we have a single source of truth for these kinds of utilities

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -701,22 +755,25 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
     )
         internal
         pure
         returns (
             bytes32
         )
     {
+        if (BYTES_TILL_TX_DATA + _txDataLength > _hashMemory.length) {
+            _hashMemory = new bytes(BYTES_TILL_TX_DATA + _txDataLength);
+        }
```
karlfloersch, line 769: We should add a test which explicitly tests this codepath. Some batch where there are a bunch of small txs and then one big one

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
karlfloersch: 🔥 `get2` ftw

karlfloersch (commented):
This looks dope! Biggest thing I think is adding tests for the more complicated memory logic

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -870,8 +946,11 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
         );
 
         // If there are some queue elements pending:
-        if (getQueueLength() - _nextQueueIndex > 0) {
-            Lib_OVMCodec.QueueElement memory nextQueueElement = getQueueElement(_nextQueueIndex);
+        if (_queueLength - _nextQueueIndex > 0) {
```
ben-chain, line 949:
```suggestion
        if (_nextContext.numSubsequentQueueElements > 0 && _queueLength - _nextQueueIndex > 0) {
```
Am I right that we can further optimize this way?  From stats below it looks like this isn't the biggest source of cost ATM, but if it sounds right to you, seems worth doing because it's this easy.

ben-chain (commented):
LGTM!  One easy optimization stood out to me that seemed worth doing, other than that I think we can ship and can approve right after the commit.  Thanks for putting in this legwork 💪

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
ben-chain: would also update the comment a line above accordingly

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
ben-chain: Oh yeah, agreed that this is worthwhile.

Additionally, how confident are we that we're correctly passing by reference here?  This would actually be a lot of cost if behind the scenes it was actually allocating new memory to pass downwards.  If it were, we could manually pass in a pointer `uint`.  I can look into this locally and see if it does anything.

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.gas.spec.ts
ben-chain: strong +1 on this; also, the first three functions here appear to be shadowed when I pulled this locally.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
ben-chain: confirming that pass by ref appears to be working; I did a manual pass by ref with same exact gas properties locally.

gakonst: Gotta make sure that you are always zeroing out the memory before you use it I think. Not sure if this happens automatically. Worth adding a test with a chainElement where there's some 0 bytes and see if they get re-used or are properly zeroed out

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
kfichter: This is actually quite a large source of cost, so I do think this is worthwhile. Although are we sure that this is valid? Do we only need to do this check when there are subsequent queue elements?

test/contracts/OVM/chain/OVM_CanonicalTransactionChain.gas.spec.ts
kfichter: Uses `core-utils` stuff now.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -406,54 +391,83 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
+        // Take a reference to the queue and its length so we don't have to keep resolving it.
+        // Length isn't going to change during the course of execution, so it's fine to simply
+        // resolve this once at the start. Saves gas.
+        iOVM_ChainStorageContainer queueRef = queue();
+        uint40 queueLength = _getQueueLength(queueRef);
+
+        // Reserve some memory to save gas on hashing later on. This is a relatively safe estimate
+        // for the average transaction size that will prevent having to resize this chunk of memory
+        // later on. Saves gas.
+        bytes memory hashMemory = new bytes((msg.data.length / totalElementsToAppend) * 2);
```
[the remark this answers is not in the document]
kfichter: We check for this earlier in the function, so we'll avoid any divide by zero:

```
        require(
            totalElementsToAppend > 0,
            "Must append at least one element."
        );
```

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
kfichter: Sounds good. Will add tests here.

tynes:
This has become stale but seems pretty important

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
maurelian: I think we should add a comment clarifying why the index is doubled.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -506,24 +520,28 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
             // The last element is a sequencer tx, therefore pull timestamp and block number from the last context.
-            timestamp = uint40(curContext.timestamp);
+            blockTimestamp = uint40(curContext.timestamp);
             blockNumber = uint40(curContext.blockNumber);
         } else {
             // The last element is a queue tx, therefore pull timestamp and block number from the queue element.
-            Lib_OVMCodec.QueueElement memory lastElement = getQueueElement(nextQueueIndex - 1);
-            timestamp = lastElement.timestamp;
+            Lib_OVMCodec.QueueElement memory lastElement = _getQueueElement(
+                nextQueueIndex - 1,
```
maurelian, line 535: can `nextQueueIndex` ever be 0?

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -438,54 +423,83 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
         for (uint32 i = 0; i < numContexts; i++) {
             BatchContext memory nextContext = _getBatchContext(i);
+
             if (i == 0) {
+                // Execute a special check for the first batch.
                 _validateFirstBatchContext(nextContext);
             }
-            _validateNextBatchContext(curContext, nextContext, nextQueueIndex);
 
+            // Execute this check on every single batch, including the first one.
```
maurelian, line 468: Appreciate these added comments.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -706,11 +724,47 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
+        iOVM_ChainStorageContainer _queueRef
+    )
+        internal
+        view
+        returns (
+            Lib_OVMCodec.QueueElement memory _element
+        )
+    {
+        (
+            bytes32 queueRoot,
```
maurelian, line 749: Why call this `queueRoot` instead of `transactionHash`?

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
maurelian: I like this much more fwiw.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -327,14 +312,14 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
+        iOVM_ChainStorageContainer queueRef = queue();
 
-        queue.push2(
+        queueRef.push2(
             transactionHash,
             timestampAndBlockNumber
         );
 
-        uint256 queueIndex = queue.length() / 2;
+        uint256 queueIndex = queueRef.length() / 2;
```
maurelian, line 325: would it be more accurate used
`uint256 queueIndex = queueRef.length() / 2 - 1`, 
rather than doing the subtraction below?

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
maurelian: it's not clear to me why we're subtracting by one here.

maurelian (commented):
Looks pretty good to me. 
I added a few comments, and am just a tad uneasy about some of the arithmetic around the `queueIndex`, especially on line 535

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
smartcontracts: New version or old version?

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
smartcontracts: I had to reread the logic but I'll do my best to explain. If we're inside of this block then `curContext.numSubsequentQueueTransactions > 0` which means that we've processed at least one queue element (or we would've thrown when we tried to process a queue element that doesn't exist). Reason why we want `nextQueueIndex - 1` is because we increment `nextQueueIndex` after processing each queue element. So the index of the last element we processed is `nextQueueIndex - 1`. I can add this as a comment in the code.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
K-Ho: @smartcontracts going to make this change unless there's any objection :)

smartcontracts: No objection.

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
smartcontracts: @ben-chain can you take another look at this?

ben-chain: Just did another playthrough; it looks like this is insufficient in the case that *all* subsequent contexts have no queue elements.

@smartcontracts If you think there are big savings, then comparing the `_finalContext` to the next pending queue element in `_validateFinalBatchContext` would make it work.  This could be done conditionally; only if the final batch context has no subsequent queue elements.

smartcontracts: Eh okay this seems sufficiently complex that we can save it for a future set of optimizations. This PR gets our gas costs down pretty hard already.

smartcontracts: I'll make a note of this in my scratch pad.

ben-chain (approved):
Re-reviewed the final changes here, and gave the whole thing a final go-through.  Looks good to ship!

PR #229, opened 2021-02-16
script: serve state dump

tynes:
## Description

A script for serving the state dump on localhost. Usage:

```bash
$ yarn build
$ yarn serve
```

This depends on python3 being installed locally

## Contributing Agreement

### DIFF
--- a/bin/serve_dump.sh
+++ b/bin/serve_dump.sh
@@ -8,9 +8,7 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" > /dev/null && pwd )"
 
-# requires python3 for now
 PYTHON=${PYTHON:-python}
-
 HOST=${HOST:-0.0.0.0}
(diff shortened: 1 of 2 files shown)


ben-chain:
Ran into troubles with this due to having my `python` be 2.7 whereas `python3` is my v3.  Was unable to simply alias to solve this.

Is this intended to be run locally or as a part of docker?  It seems like implementing this in JS would decrease the local dependencies required.

tynes:
I'll update this to work better. Not sure why python2 is still the default python on osx, its been sunset for over a year at this point

tynes:
Note that there is a way to get around this with how I wrote the script

```
$ PYTHON=python3 ./scripts/serve_dump.sh
```

tynes:
This should now work for python2 or python3

karlfloersch (approved):
LGTM!

PR #233, opened 2021-02-18
Rebuild lockfile

ben-chain:
## Description
Rebuilds lockfile so that BigNumber implementations are consistent between ethers stuff.  See: https://github.com/ethers-io/ethers.js/issues/1293

## Contributing Agreement


kfichter:
Meh we need to drop support for Node 11 anyway...

tynes:
No need to support odd versions as even versions are the ones that are considered stable

tynes (approved):

PR #244, opened 2021-02-23
feat/641/initNonZero: Use max values instead of zeros in reset context

maurelian:
## Description

Sets most values context values to non-zero by default. 

Todo:
- [ ] Similar treatment for the reentrancy guard.

### Security consideration:

There is a danger that one of these values is read _before_ it is set. This was the case with the `ovmGasRefund` value, so we need to ensure there are no other such cases.

### Fixes
- Fixes https://github.com/ethereum-optimism/roadmap/issues/641

## Contributing Agreement


contracts/optimistic-ethereum/iOVM/execution/iOVM_ExecutionManager.sol
```diff
@@ -11,6 +11,7 @@ interface iOVM_ExecutionManager {
      *********/
 
     enum RevertFlag {
+        // UNUSED, // Save gas by not setting a storage slot to zero
```
maurelian, line 14: Added this here for discussion. 
Currently it breaks all the revert flag tests, but they shouldn't be too hard to fix once I know what's going on in there.

ben-chain: ah yes, there is an off-chain revert flag enum somewhere in like `test-helpers` that will need to be updated.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1789,20 +1796,20 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-        transactionContext.ovmGASLIMIT = 0;
-        transactionContext.ovmTXGASLIMIT = 0;
+        transactionContext.ovmL1TXORIGIN = MAX_ADDRESS;
+        transactionContext.ovmTIMESTAMP = type(uint256).max;
+        transactionContext.ovmNUMBER = type(uint256).max;
+        transactionContext.ovmGASLIMIT = type(uint256).max;
+        transactionContext.ovmTXGASLIMIT = type(uint256).max;
         transactionContext.ovmL1QUEUEORIGIN = Lib_OVMCodec.QueueOrigin.SEQUENCER_QUEUE;
 
         transactionRecord.ovmGasRefund = 0;
```
maurelian, line 1806: This is the only value I didn't set to `type(uint256).max`. The reason being that several test cases failed, because it seems that in those cases (mostly static call violations), the default gas refund value is used.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -205,7 +207,8 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         _resetContext();
 
         // Reset the ovmStateManager.
-        ovmStateManager = iOVM_StateManager(address(0));
+        // @todo: should we move this into reset context?
```
ben-chain, line 210: seems good to me!

ben-chain (dismissed):

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -205,7 +211,8 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+        ovmStateManager = iOVM_StateManager(MAX_ADDRESS);
```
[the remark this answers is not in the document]
maurelian: Let's discuss because now that I look more closely, that would also have the effect of resetting the SM up here:

https://github.com/ethereum-optimism/contracts/blob/2a29fe559f3a11ce8d7f5c9b6c7e8990c0569f89/contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol#L185-L191

I'm also just a bit unclear about the purpose of resetting the context there, even prior to this change.

test/contracts/OVM/execution/OVM_ExecutionManager/ovmREVERT.spec.ts
```diff
@@ -16,6 +16,9 @@ const test_ovmREVERT: TestDefinition = {
       messageRecord: {
         nuisanceGasLeft: OVM_TX_GAS_LIMIT,
       },
+      transactionRecord: {
+        ovmGasRefund: 0,
+      },
```
maurelian, line 21: OK, we fixed the tests, but I think we also need to manually verify that there is no way to reach a pre-state with non-zero `ovmGasRefund`. I've been walking through the relevant code paths, but I think we should discuss together.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1767,43 +1770,61 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+        transactionContext.ovmGASLIMIT = gasMeterConfig.maxGasPerQueuePerEpoch;
         transactionContext.ovmTXGASLIMIT = _transaction.gasLimit;
-        transactionContext.ovmL1QUEUEORIGIN = _transaction.l1QueueOrigin;
         transactionContext.ovmL1TXORIGIN = _transaction.l1TxOrigin;
-        transactionContext.ovmGASLIMIT = gasMeterConfig.maxGasPerQueuePerEpoch;
+
+        transactionRecord.ovmGasRefund = 0;
+
+        messageContext.ovmCALLER = address(-1);
+        messageContext.ovmADDRESS = address(-1);
```
ben-chain, line 1792: I think that this is not strictly necessary, but I like the redundancy

contracts/optimistic-ethereum/libraries/codec/Lib_OVMCodec.sol
```diff
@@ -37,6 +37,7 @@ library Lib_OVMCodec {
     }
 
     enum QueueOrigin {
+        UNUSED, // Save gas by not setting a storage slot to zero
```
ben-chain, line 40: This one actually has the chance of requiring downstream changes due to how this enum is handled in geth.  I think we may want to take the efficiency loss to keep this as it was.  CC @karlfloersch  do you know?

ben-chain (commented):
I think we will want to revert one change just for not having a headache, but LGTM otherwise.  Pinged Karl for an answer to that q.

test/helpers/test-runner/test-runner.ts
```diff
@@ -89,7 +89,7 @@ export class ExecutionManagerTestRunner {
     )),
       (test.postState = test.postState || {})
 
-    describe(`OVM_ExecutionManager Test: ${test.name}`, () => {
+    describe.only(`OVM_ExecutionManager Test: ${test.name}`, () => {
```
ben-chain, line 92:
```suggestion
    describe(`OVM_ExecutionManager Test: ${test.name}`, () => {
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -158,7 +158,9 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         override
         public
     {
-        require(transactionContext.ovmNUMBER == 0, "Only be callable at the start of a transaction");
+        // Prevent reentrancy to this function by requiring that the transaction context has not yet been
+        // initialized
+        require(transactionContext.ovmNUMBER == type(uint256).max, "Only callable as the start of a transaction");
```
kfichter, line 163:
```suggestion
        require(transactionContext.ovmNUMBER == type(uint256).max, "OVM_ExecutionManager: Run can only be called at the start of a transaction.");
```

maurelian: Do we have a max line length? I've been feeling like we should. Generally I try not to go over 120.

contracts/optimistic-ethereum/libraries/codec/Lib_OVMCodec.sol
ben-chain: Talked to the team, def not worth the risk of missing an external dependency -- let's revert this one.

kfichter:
Would be great if we could add tests that show the amount of gas saved with these optimizations

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
maurelian: Discussed offline, we are keeping this, as well as setting them to `address(0)` in `_initContext()`.

.github/workflows/integration.yml
```diff
@@ -28,9 +28,13 @@ jobs:
               GITHUB_HEAD_REF=${GITHUB_REF##*/}
           fi
           GIT_COMMIT=$(git rev-parse HEAD | head -c 8)
-          REMOTE="https://github.com/${{ github.repository }}"
+          REMOTE=${{ github.event.pull_request.head.repo.html_url }}
```
maurelian, line 31: I'm confused about why this is showing up in the diff. The current state of master looks like the right hand side, not the left: 

https://github.com/ethereum-optimism/contracts/blob/master/.github/workflows/integration.yml#L31-L37

maurelian:
This is ready for final review, once #263 has been merged.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
maurelian: If we ever revive this, I'm now of the opinion that something like `0x000000000000000000000000000000000000000000000000000000000000dead` would be better. I don't like being within an 'off-by-one' error of zero.

maurelian:
I think this is obsoleted by other work. Closing for now.

PR #245, opened 2021-02-24
wip: verify script

tynes:
## Description

WIP script to verify all of the contracts

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

- [ ] I have read and understood the [Optimism Contributing Guide and Code of Conduct](https://github.com/ethereum-optimism/optimism-monorepo/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

PR #249, opened 2021-02-24
feat: Add support for hardhat-deploy

### LINKED ISSUE #116: Add nonce manager for quick deploys
**Is your feature request related to a problem? Please describe.**
As it stands all of our transactions are sent without waiting for receipts & without hardcoding the nonce. This means we often fail during "real" deployments (to chains other than ganache).

**Describe the solution you'd like**
Add a function `sendTransaction(tx)` which handles incrementing the nonce of each tx. This will allow us to send all of our transactions upfront & just wait until they all get mined vs waiting around for receipts.

**Describe alternatives you've considered**
Wait for receipts

### LINKED ISSUE #256: Add progress bar to contract deployment
I ran deploy, and while I see the contracts going on to the chain, I don't see anything in the console, been hanging for about 10 minutes now.

smartcontracts:
## Description
Adding new support for `hardhat-deploy` as an alternative to our bespoke deployment system. `hardhat-deploy` is quite nice and generates very clean artifacts. Next step should be to remove the old deployment system when possible. Though we might have to do something about generating state dumps.

fixes #256
fixes #116

## Contributing Agreement


contracts/optimistic-ethereum/libraries/resolver/Lib_AddressResolver.sol
```diff
@@ -13,7 +13,7 @@ abstract contract Lib_AddressResolver {
      * Contract Variables: Contract References *
      *******************************************/
 
-    Lib_AddressManager internal libAddressManager;
+    Lib_AddressManager public libAddressManager;
```
smartcontracts, line 16: This is actually very security critical. Some of our (annoying) proxies require that we call `initialize()` outside of the constructor. Making this public allows us to check that the address has been set correctly when `initialize()` is called.

src/hardhat-deploy-ethers.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // Now reset Object.defineProperty
+  Object.defineProperty = def
+
+  for (const fnName of Object.keys(contract.functions)) {
+    const fn = contract[fnName].bind(contract)
+    ;(contract as any)[fnName] = async (...args: any) => {
+      const result = await fn(...args)
+      if (typeof result === 'object' && typeof result.wait === 'function') {
+        await result.wait()
```
smartcontracts, line 52: Call `.wait()` here so we don't have to do it every single time we want to interact with a contract.

tslint.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@eth-optimism/dev/tslint.json",
   "rules": {
     "array-type": false,
-    "class-name": false
+    "class-name": false,
+    "prefer-conditional-expression": false
```
smartcontracts, line 6: I think conditional expressions are pretty dumb

src/hardhat-deploy-ethers.ts
```diff
@@ -0,0 +1,59 @@
...
+  hre: HardhatRuntimeEnvironment,
+  name: string,
+  options: {
+    iface?: string
+    signerOrProvider?: Signer | Provider | string
+  } = {}
+): Promise<Contract> => {
+  const deployed = await hre.deployments.get(name)
+
+  await hre.ethers.provider.waitForTransaction(deployed.receipt.transactionHash)
```
smartcontracts, line 17: Just make sure we're interacting with a contract that actually exists. Not sure if this is strictly necessary but seems better to be on the safe side.

deploy/Proxy__OVM_L1CrossDomainMessenger.deploy.ts
```diff
@@ -0,0 +1,61 @@
...
+      signerOrProvider: deployer,
+      iface: 'OVM_L1CrossDomainMessenger',
+    }
+  )
+
+  await Proxy__OVM_L1CrossDomainMessenger.initialize(Lib_AddressManager.address)
+
+  const libAddressManager = await Proxy__OVM_L1CrossDomainMessenger.libAddressManager()
+  if (libAddressManager !== Lib_AddressManager.address) {
+    throw new Error(
```
smartcontracts, line 43: We can do some cool post-deploy verification checks now!

tslint.json
gakonst: Agree, not a fan of this syntax except for very simple cases

contracts/optimistic-ethereum/libraries/resolver/Lib_AddressResolver.sol
tynes: This needs to be in a separate PR so that this one only touches hardhat deploy related things

tslint.json
tynes: Can this go into `@eth-optimism/dev` and then we update the dep in another PR?

tynes:
I'd really like it if you could add an `index.js` that does something like:

```js
const { LedgerSigner } = require('@ethersproject/hardware-wallets');

extendEnvironment((hre) => {
   const hdpath = ethers.utils.defaultPath || process.env.HD_PATH
   hre.ledger = new LedgerSigner(hre.provider, 'default', hdpath);
})
```
Ideally there is a flag `--ledger` and it just overwrites the `hre.signer` so that we don't need logic around using `hre.ledger` in each implementation of the contract deployment

smartcontracts:
@tynes I started a conversation about hardware wallet support over here https://github.com/wighawag/hardhat-deploy/issues/88. I'm going to try experimenting with this today (assuming someone has a ledger that I can play with...)

contracts/optimistic-ethereum/libraries/resolver/Lib_AddressResolver.sol
smartcontracts: Ok I'll make a quick PR for this.

smartcontracts: PR is open here: https://github.com/ethereum-optimism/contracts/pull/352

smartcontracts: Done.

tslint.json
smartcontracts: Change to `tslint.json` was included in this PR because tslint was complaining about a file and wanted to make the file *way* uglier.

hardhat.config.ts
```diff
@@ -8,14 +8,19 @@ import {
...
 
 const config: HardhatUserConfig = {
   networks: {
     hardhat: {
       accounts: DEFAULT_ACCOUNTS_HARDHAT,
       blockGasLimit: RUN_OVM_TEST_GAS * 2,
+      live: false,
+      saveDeployments: false,
+      tags: ['test', 'local'],
     },
```
gakonst, line 24: Should probably add new mainnet/rinkeby/kovan etc. in the networks for our actual deployments.

src/hardhat-deploy-ethers.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  // TODO: Cache these 2 across calls?
+  const { deployer } = await hre.getNamedAccounts()
+  const Lib_AddressManager = await getDeployedContract(
+    hre,
+    'Lib_AddressManager',
+    {
+      signerOrProvider: deployer,
+    }
+  )
```
gakonst, line 28: This is called twice in each script now, once in the external deploy script and once here.

src/hardhat-deploy-ethers.ts
```diff
@@ -0,0 +1,94 @@
...
+    prop.writable = true
+    return def(obj, propName, prop)
+  }
+
+  const contract = new Contract(deployed.address, iface, signerOrProvider)
+
+  // Now reset Object.defineProperty
+  Object.defineProperty = def
+
+  for (const fnName of Object.keys(contract.functions)) {
```
gakonst, line 82:
```suggestion
  // Override each function call to also `.wait()` so as to simplify the deploy scripts' syntax.
  for (const fnName of Object.keys(contract.functions)) {
```

src/hardhat-deploy-ethers.ts
gakonst: I don't think this is a great change because it means that we'll have to "block" on the address manager registration being mined each time, resulting in slow deployments. OK for now, but let's try to remove in the future?

gakonst (dismissed):
LGTM, some minor comments in case you want to address before merging, otherwise go for it.

maurelian (approved):
LGTM. 
One thing I noticed is that there are not scripts for bridge contracts. Should there be a ticket to migrate them to HH?

snario:
@smartcontracts can you provide an update on this pull request? What's your intention for next steps? Given that it has been approved and the tests pass, I'm inclined to merge it, but I understand you've been looking at other deploy options?

smartcontracts:
Also makes #245 obsolete

src/hardhat-deploy-ethers.ts
```diff
@@ -0,0 +1,131 @@
...
+  let signerOrProvider: Signer | Provider = hre.ethers.provider
+  if (options.signerOrProvider) {
+    if (typeof options.signerOrProvider === 'string') {
+      signerOrProvider = hre.ethers.provider.getSigner(options.signerOrProvider)
+    } else {
+      signerOrProvider = options.signerOrProvider
+    }
+  }
+
+  // Temporarily override Object.defineProperty to bypass ether's object protection.
```
tynes, line 106: Safe code is safe

src/hardhat-deploy-ethers.ts
```diff
@@ -0,0 +1,131 @@
...
+    const factory = await hre.ethers.getContractFactory(options.iface)
+    iface = factory.interface
+  }
+
+  let signerOrProvider: Signer | Provider = hre.ethers.provider
+  if (options.signerOrProvider) {
+    if (typeof options.signerOrProvider === 'string') {
+      signerOrProvider = hre.ethers.provider.getSigner(options.signerOrProvider)
+    } else {
+      signerOrProvider = options.signerOrProvider
```
tynes, line 102: Will this allow for the `LedgerSigner`?

src/predeploys.ts
```diff
@@ -0,0 +1,12 @@
+export const predeploys = {
```
tynes, line 1: A nice to have imo would be to export these as contract objects that are already connected to the correct address

hardhat.config.ts
```diff
@@ -37,6 +46,67 @@ const config: HardhatUserConfig = {
...
+    deploy: './deploy',
+    deployments: './deployments',
+  },
+  namedAccounts: {
+    deployer: {
+      default: 0,
+    },
+  },
+}
+
```
tynes, line 59: Why specify a key and url for each network instead of just `DEPLOYER_KEY` `RPC_URL` and `NETWORK`?

bin/deploy.ts
```diff
@@ -0,0 +1,35 @@
...
+    emMaxGasPerQueuePerEpoch: process.env.MAX_GAS_PER_QUEUE_PER_EPOCH,
+    emSecondsPerEpoch: process.env.SECONDS_PER_EPOCH,
+    emOvmChainId: process.env.CHAIN_ID,
+    sccFraudProofWindow: process.env.FRAUD_PROOF_WINDOW_SECONDS,
+    sccSequencerPublishWindow: process.env.SEQUENCER_PUBLISH_WINDOW_SECONDS,
+    ovmSequencerAddress: process.env.SEQUENCER_ADDRESS,
+    ovmRelayerAddress: process.env.RELAYER_ADDRESS,
+  })
+}
+
```
tynes, line 27: Small nit as `then` needs to be deprecated

```js
;(async () => {

})().catch(err => {

})

```

hardhat.config.ts
smartcontracts: good point

src/hardhat-deploy-ethers.ts
smartcontracts: Yep, exactly

smartcontracts: Although realistically I still think it's better to do these deployments from hot wallets and have verification once the deployment is finished, then transfer ownership over to the multisig.

bin/deploy.ts
smartcontracts: Was following https://hardhat.org/guides/scripts.html. And the ;(...)() syntax was giving me a lot of headache with the linter so I went with this.

src/predeploys.ts
smartcontracts: Yeah, will leave that for a future PR.

deploy/000-Lib_AddressManager.deploy.ts
```diff
@@ -0,0 +1,51 @@
...
+  await registerAddress({
+    hre,
+    name: 'OVM_Sequencer',
+    address: (hre as any).deployConfig.ovmSequencerAddress,
+  })
+
+  await registerAddress({
+    hre,
+    name: 'OVM_Proposer',
+    address: (hre as any).deployConfig.ovmSequencerAddress,
```
ben-chain, line 39:
```suggestion
    address: (hre as any).deployConfig.ovmProposerAddress,
```
(even if these are the same now might as well(

deploy/007-OVM_L1CrossDomainMessenger.deploy.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const OVM_L1CrossDomainMessenger = await getDeployedContract(
+    hre,
+    'OVM_L1CrossDomainMessenger',
+    {
+      signerOrProvider: deployer,
+    }
+  )
+
+  await OVM_L1CrossDomainMessenger.initialize(Lib_AddressManager.address)
```
ben-chain, line 37: can we add a clarifying comment here that this is technically not required?  Just for auditability

deploy/015-OVM_L1ETHGateway.deploy.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  if (!result.newlyDeployed) {
+    return
+  }
+
+  const OVM_L1ETHGateway = await getDeployedContract(hre, 'OVM_L1ETHGateway', {
+    signerOrProvider: deployer,
+  })
+
+  await OVM_L1ETHGateway.initialize(
```
ben-chain, line 34: ditto above -- comment on redundant initialization for auditability :)

ben-chain (changes requested):
Requested a few nit changes, but looking great!

ben-chain (approved):
LGTM pending CI!

ben-chain (approved):
👍

PR #253, opened 2021-02-26
Feat/649/cached safety checking

maurelian:
## Description

Increase gas efficiency on L2 by adding a new contract to cache known safe bytecode.

https://github.com/ethereum-optimism/roadmap/issues/649


ben-chain:
The main thing that needs to be sorted out around this PR are:
- [ ] Choose between this approach and https://github.com/ethereum-optimism/contracts/pull/310
- [ ]  Fix how this will work in the fraud proofs.  Right now, the safety checking logic has the potential to bypass the `INVALID_STATE_ACCESS` logic in place with `_handleExternalMessage`.

maurelian:
> * [ ]   Fix how this will work in the fraud proofs.  Right now, the safety checking logic has the potential to bypass the `INVALID_STATE_ACCESS` logic in place with `_handleExternalMessage`.

To clarify somewhat: 
Unlike the safety checker, the safety cache is stateful, meaning an inclusion proof needs to be provided for it prior to fraud proving on L1. 

One way to handle this would be a similar approach to how the `OVM_DeployerWhitelist` is called using `ovmCall` which includes all the logic necessary to require an inclusion proof.

maurelian:
This PR builds on top of code that has been removed. 
We should start from scratch on a new branch.

PR #261, opened 2021-03-02
[Feat] Autopublish with new version

annieke:
## Description

This change enables autopublishing from updating the version in `package.json`.

### How it works

When there are no changes to the version in `package.json`, the `salsify/action-detect-and-tag-new-version@v2` runs but does not product a new tag, so the subsequent jobs (release and publish) do not run. [Example](https://github.com/ethereum-optimism/service-base/actions/runs/612366930) in service-base.

When the version # in `package.json` is updated, the action creates a tag corresponding to the version number (so we should make sure this matches what we want in release!), which then triggers the release action and publish action. [Example of this case](https://github.com/ethereum-optimism/service-base/runs/2008914296?check_suite_focus=true) in service-base.

## Questions
- 
-
-

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement


.github/workflows/tag-release.yml
```diff
@@ -1,36 +1,68 @@
-name: Create Releases on Tags
+name: Auto tag-release-publish
 
 on:
   push:
-    tags:
-      - 'v*'
+    branches:
+      - master
```
tynes, line 6: Is this going to tag a new release on each merge to master?

tynes:
Could you describe exactly what this is supposed to do so that I can verify that it works as expected?

.github/workflows/tag-release.yml
annieke: nope! it will only tag a new release when the version in `package.json` is updated

annieke:
@tynes just updated the PR description

tynes (approved):

.github/workflows/tag-release.yml
```diff
@@ -90,6 +90,8 @@ jobs:
     name: Commit State Dump
     needs: docker
     runs-on: ubuntu-latest
+    needs: tag
+    if: needs.tag.outputs.tag_nam
```
snario, line 94:
```suggestion
    if: needs.tag.outputs.tag_name
```

.github/workflows/tag-release.yml
```diff
@@ -1,36 +1,68 @@
...
+    needs: tag
+    if: needs.tag.outputs.tag_name
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: yarn
+      - run: yarn build
```
snario, line 50: This task (i.e., `yarn build`) is also run inside the docker container in the `docker` job. It might not be a priority now, but you could theoretically fetch the built artifacts and push those to `npm` as opposed to building them twice. That would lower the total time this workflow takes.

annieke: ooo this is a good idea! i'll make an tech debt or onboarding ticket for this, best place for that @karlfloersch ?

annieke: made it here: https://github.com/ethereum-optimism/roadmap/issues/725

tynes: Good job making a ticket for it

snario:
Two more comments that came to mind this morning.

1. This does change the behaviour from tagging using `git` tags to using the `version` flag of `package.json`. Both things require manual effort by the developer to decide to tag so it's a similar workflow which is nice. However, it does prevent using git tags strategically for things like canary or alpha releases for code that is not yet on `master`. Has that been considered?

2. It would be worthwhile adding a section to the `README` to explain a little bit about how this works and what the expected workflow by the developer is. For example some questions worth answering are:
    - How I publish packages to `npm`?
    - What is the accepted standard deciding when to bump the packages version?

annieke:
@snario both good points! 

1. Thought about this a bit - I'm hoping that for the `on: branches:` section of the file we can specify branch names that follow a certain regex or just name them canary/alpha, and it SHOULD work. Definitely down to make a ticket for that as improvement since we thought about beta releases before.
2. Ooo this is a good idea! I don't have a good idea of the second question yet, so will try to find out and add this as a followup PR.

maurelian (approved):
LGTM

PR #269, opened 2021-03-02
Commit state dumps to regenesis repo for new tags

snario:
Creates a commit on https://github.com/ethereum-optimism/regenesis when a new Docker image is successfully built and pushed to Docker hub that contains the state dump, labelled using the same tag.

## Description
This is implemented as a new job on the [Create Releases on Tags](https://github.com/ethereum-optimism/contracts/actions/workflows/tag-release.yml) Github Action to take advantage of the fact that (1) that workflow only runs when a new tag is created and (2) it already builds the state dump in the `docker` job so we can re-use it.

The `docker` job now stores the state dump as an artifacts on the Github Action: https://github.com/ethereum-optimism/contracts/blob/34724a1e553623f2f68ff5adb6728d0c4f7f1abb/.github/workflows/tag-release.yml#L48-L52

The `git` job will read that artifact and create a commit on `regenesis`: https://github.com/ethereum-optimism/contracts/blob/d946f8c2429bbe539384af67f8225310196d4804/.github/workflows/tag-release.yml#L73-L80

## Questions / TODO
- ~I'll need to add a secret called `REGENESIS_COMMIT_PAT`. Using ssh keys [does not work ](https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#creating-pull-requests-in-a-remote-repository)since this is creating a pull request in a separate repository.~
- ~Right now the file is just stored in the root of the repo; we probably want to put them inside a folder called `tags`? Would need to open a PR in `regenesis` to add that folder.~

### Fixes
- Fixes https://github.com/ethereum-optimism/roadmap/issues/694

## Contributing Agreement

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -42,5 +42,39 @@ jobs:
           ./build.sh -s deployer -b ${{ steps.get_version.outputs.VERSION }}
           docker push ethereumoptimism/deployer:${{ steps.get_version.outputs.VERSION }}
 
+      - name: Extract State Dump from Deployer Image
+        run: docker run --rm --entrypoint cat ethereumoptimism/deployer:${{ steps.get_version.outputs.VERSION }} /opt/contracts/build/dumps/state-dump.latest.json > ${{ steps.get_version.outputs.VERSION }}.json
+
+      - name: Store state dump artifact for git job
+        uses: actions/upload-artifact@v2
+        with:
+          name: state-dump-artifact
+          path: ${{ steps.get_version.outputs.VERSION }}.json
+
       - name: Logout of DockerHub
         run: docker logout
+
+  git:
+    name: Commit State Dump
+    needs: docker
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: ethereum-optimism/regenesis
+          ssh-key: ${{ secrets.REGENESIS_DEPLOY_KEY }}
+
+      - name: Download State Dump Artifact
+        uses: actions/download-artifact@v2
+        with:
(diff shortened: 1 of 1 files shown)


snario:
AFAICT this is currently blocked on either:
- Deciding to store these state dumps in this same repository (in which case we don't need a new deploy key)
- Provisioning a new deploy key for the `regenesis` repo which can be used in this repositories Github Actions as a secret

.github/workflows/tag-release.yml
```diff
@@ -42,5 +42,41 @@ jobs:
...
+  git:
+    name: Commit State Dump
+    needs: docker
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: ethereum-optimism/regenesis
+          ssh-key: ${{ secrets.INTEGRATION_PUSH_ACTION }}
```
snario, line 66: To be replaced with a deploy key from `regenesis`

snario: Nevermind — changed `INTEGRATION_PUSH_ACTION` to be enabled for `regenesis`!

snario:
Working example of a commit by @actions-user [here](https://github.com/ethereum-optimism/regenesis/commit/a1f229d946c883d21593bc533f0c209b232e960e) and the related passing [job](https://github.com/ethereum-optimism/contracts/runs/2016457652?check_suite_focus=true).

tynes (approved):
Looks good to me!

snario:
First successful tag on `master`: https://github.com/ethereum-optimism/regenesis/commit/e9afa665800e330966a0988b1b1d77f1167cbe2d

PR #273, opened 2021-03-03
Add receive() to OVM_L1ETHGateway so you can send ETH to it

geohot:


geohot:
Also added fallback function for deposit.

tonysosa-dev (approved):

geohot:
Unclear why tests are failing. It would also be a good idea to add a test for the receive behavior.

tynes:
Its failing due to a compiler error and a problem with the ovm compiler plugin causes it to surface in this way
```
TypeError: Cannot use 'in' operator to search for '@openzeppelin/contracts/math/SafeMath.sol' in undefined
    at OverriddenTaskDefinition.config_1.subtask [as _action] (/home/runner/work/contracts/contracts/node_modules/@eth-optimism/plugins/src/hardhat/compiler/index.ts:154:19)
 ```

gakonst:
If you remove the optimism compiler from the hardhat plugins you can see the real error. For some reason the hardhat plugin swallows the error.

maurelian:
The solc error message is `Error: Only one receive function is allowed`, because there is already a receive function at the bottom of the file: 

https://github.com/ethereum-optimism/contracts/blob/6c895e2bb90dcad09b6bb4282333c5f162dcc7fb/contracts/optimistic-ethereum/OVM/bridge/tokens/OVM_L1ETHGateway.sol#L161-L166

Personally I think it's a bit cleaner to require calling the `deposit()` function.

tynes:
> IMO, it's a bit cleaner to require calling the deposit() function.

I think its a bit more convenient for users to be able to send ETH directly to the contract and have it be deposited. Then simple wallets will be able to be used to deposit as well

ben-chain:
I think that the argument makes sense.  Also even if we don't do it, someone could just build a wrapper contract. I'll see if I can get this merged in tomorrow.

maurelian:
Couldn't push to this branch, so I opened a new PR  (#311) with the fix.

PR #285, opened 2021-03-03
Update OVM_L1ERC20Gateway.sol

tonysosa-dev:
Respectfully submitted 

## Contributing Agreement

- [x] I have read and understood the [Optimism Contributing Guide and Code of Conduct](https://github.com/ethereum-optimism/optimism-monorepo/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

PR #286, opened 2021-03-03
Update OVM_L1ERC20Gateway.sol

tonysosa-dev:
Period added extra bit returning error then red x in view code field. 

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/bridge/tokens/OVM_L1ERC20Gateway.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/tokens/OVM_L1ERC20Gateway.sol
@@ -82,7 +82,7 @@ contract OVM_L1ERC20Gateway is Abs_L1TokenGateway {
 
     /**
      * @dev When a withdrawal is finalized on L1, the L1 Gateway
-     * transfers the funds to the withdrawer.
+     * transfers the funds to the withdrawer
      *
      * @param _to L1 address that the ERC20 is being withdrawn to
      * @param _amount Amount of ERC20 to send


kfichter:
Thanks for the contribution again :-)

> Period added extra bit returning error then red x in view code field.

Could you elaborate on this? Not sure I understand.

tonysosa-dev:
Update OVM_L1ERC20Gateway.sol #285 showed an error(red x)when viewing the newly committed code.  This was because I included an extra bit in the form of a period where I should not have   withdrawer. 
instead of                           withdrawer 
✅

kfichter:
I think it should be ok to keep the period here.

tonysosa-dev:
I’m new to smart contracts and probably mistaken here lol, but from what I understand they are compiled into byte code, and from the screen shot I attached, I can see there are the 10 original bytes in withdraer.      My correction of withdrawer. of course contains 11.  Would this cause an error in recompiling and thus the red x now on the side of my contribution in the ethereum-optimism/contracts GitHub page?  If so I have offered a correction, because however small it may seem, it would feel great to have properly contributed thus having a green check indicating.  Thank you and your team for your time.  Any future discourse on the revolutionary OVM, I would feel very fortunate to participate.  Thank you again. On Wed, 3 Mar 2021 at 3:43 PM Kelvin wrote:
I think it should be ok to keep the period here.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/ethereum-optimism/contracts/pull/286#issuecomment-790081055",
"url": "https://github.com/ethereum-optimism/contracts/pull/286#issuecomment-790081055",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

PR #300, opened 2021-03-08
feat: Use standard RLP transactions

smartcontracts:
## Description
Updates our sequencer entrypoint and ECDSA contracts to use the standard RLP encoding that we're using in geth now.

### Fixes
- Fixes roadmap#754

## Contributing Agreement


contracts/optimistic-ethereum/OVM/accounts/OVM_ProxyEOA.sol
```diff
@@ -3,8 +3,6 @@ pragma solidity >0.5.0 <0.8.0;
 
 /* Library Imports */
 import { Lib_Bytes32Utils } from "../../libraries/utils/Lib_Bytes32Utils.sol";
-import { Lib_OVMCodec } from "../../libraries/codec/Lib_OVMCodec.sol";
```
smartcontracts, line 6: Files were unused, I removed them.

contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
```diff
@@ -2,9 +2,7 @@
 pragma solidity >0.5.0 <0.8.0;
 
 /* Library Imports */
-import { Lib_BytesUtils } from "../../libraries/utils/Lib_BytesUtils.sol";
-import { Lib_OVMCodec } from "../../libraries/codec/Lib_OVMCodec.sol";
-import { Lib_ECDSAUtils } from "../../libraries/utils/Lib_ECDSAUtils.sol";
+import { Lib_EIP155Tx } from "../../libraries/codec/Lib_EIP155Tx.sol";
```
kfichter, line 5: This file gets way simpler.

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
```diff
@@ -0,0 +1,111 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >0.5.0 <0.9.0;
```
kfichter, line 2: Pulled decoding logic into this file to make review easier. It's mostly the same as what was originally in OVM codec. We can confirm this with some careful review.

contracts/optimistic-ethereum/libraries/utils/Lib_ECDSAUtils.sol
```diff
@@ -1,98 +0,0 @@
-// SPDX-License-Identifier: MIT
```
smartcontracts, line 1: File is no longer needed.

contracts/optimistic-ethereum/mockOVM/accounts/mockOVM_ECDSAContractAccount.sol
```diff
@@ -1,95 +0,0 @@
-// SPDX-License-Identifier: MIT
```
smartcontracts, line 1: We don't use this in geth and we shouldn't use this in `ethereumjs-vm`.

test/data/json/libraries/codec/Lib_OVMCodec.test.json
```diff
@@ -1,22 +1,4 @@
 {
     "tests": {
-        "decompressEIP155Transaction": {
```
smartcontracts, line 3: Tests move to new `eip155tx` library.

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
```diff
@@ -0,0 +1,119 @@
...
+pragma experimental ABIEncoderV2;
+
+/* Library Imports */
+import { Lib_RLPReader } from "../rlp/Lib_RLPReader.sol";
+import { Lib_RLPWriter } from "../rlp/Lib_RLPWriter.sol";
+
+import { console } from "hardhat/console.sol";
+
+library Lib_EIP155Tx {
+    struct EIP155Tx {
```
tynes, line 12: This could be reordered to fit in less words right?

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
```diff
@@ -0,0 +1,119 @@
...
+        uint256 nonce;
+        uint256 gasPrice;
+        uint256 gasLimit;
+        address to;
+        uint256 value;
+        bytes data;
+        uint8 recoveryParam;
+        uint8 v;
```
tynes, line 20: `v` will overflow here I think

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
```diff
@@ -0,0 +1,119 @@
...
+
+library Lib_EIP155Tx {
+    struct EIP155Tx {
+        uint256 nonce;
+        uint256 gasPrice;
+        uint256 gasLimit;
+        address to;
+        uint256 value;
+        bytes data;
+        uint8 recoveryParam;
```
tynes, line 19: This is basically just caching the recovery param so that it is only computed once instead of needing to compute it when calling `sender` which is only called once and is always called. Might save gas to use it depending on how well packed this struct can be made

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
tynes: I guess depending on what the chainid is

tynes:
Overall good changes. The abi is changing, we need to make sure to not break everything when we publish these changes. Also  we dont have hardhat deploy so there are no deployment artifacts in this repo so we need to be mindful about this change.

contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
```diff
@@ -19,109 +17,39 @@ import { Lib_SafeExecutionManagerWrapper } from "../../libraries/wrappers/Lib_Sa
...
-     * calldata[75:95]: transaction target address
-     * calldata[95:XX]: transaction data
-     */
     fallback()
         external
     {
-        TransactionType transactionType = _getTransactionType(Lib_BytesUtils.toUint8(msg.data, 0));
-
-        bytes32 r = Lib_BytesUtils.toBytes32(Lib_BytesUtils.slice(msg.data, 1, 32));
-        bytes32 s = Lib_BytesUtils.toBytes32(Lib_BytesUtils.slice(msg.data, 33, 32));
```
tynes: Removes the calls to `slice` which is good

tynes:
The calldata size of using a RLP encoded tx is about ~250 gas more than using the custom serialization. Trading a bit of cost for simplicity, but its possible that the gas usage was decreased elsewhere in these changes. We really need automated gas metering in CI

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
```diff
@@ -6,11 +6,12 @@ pragma experimental ABIEncoderV2;
 import { iOVM_ECDSAContractAccount } from "../../iOVM/accounts/iOVM_ECDSAContractAccount.sol";
 
 /* Library Imports */
-import { Lib_OVMCodec } from "../../libraries/codec/Lib_OVMCodec.sol";
-import { Lib_ECDSAUtils } from "../../libraries/utils/Lib_ECDSAUtils.sol";
+import { Lib_EIP155Tx } from "../../libraries/codec/Lib_EIP155Tx.sol";
 import { Lib_SafeExecutionManagerWrapper } from "../../libraries/wrappers/Lib_SafeExecutionManagerWrapper.sol";
 import { Lib_SafeMathWrapper } from "../../libraries/wrappers/Lib_SafeMathWrapper.sol";
 
+import { console } from "hardhat/console.sol";
```
tynes, line 13: I know this is WIP, but just noting to remove this

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
kfichter: Hmmm. I looked into this. It appears that the tight packing only applies if we plan to store this struct. Since we're only using this in memory the order doesn't matter. Source here: https://docs.soliditylang.org/en/v0.8.2/internals/layout_in_memory.html#example-for-difference-in-struct-layout

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
kfichter: Right. Given a chain ID, the recovery param will never change. So it seems a bit easier to compute this once (we *do* actually end up computing this in two different places otherwise).

contracts/optimistic-ethereum/libraries/codec/Lib_EIP155Tx.sol
kfichter: Overflow doesn't matter for the `v` parameter. I'll add a comment that explains why.

kfichter: Added some more detailed comments.

contracts/optimistic-ethereum/libraries/rlp/Lib_RLPWriter.sol
```diff
@@ -92,6 +92,23 @@ library Lib_RLPWriter {
...
+    function writeBytes32(
+        bytes32 _in
+    )
+        internal
+        pure
+        returns (
+            bytes memory _out
+        )
+    {
+        return writeBytes(abi.encodePacked(_in));
```
kfichter, line 109: `bytes32` isn't treated any differently from `bytes` when it comes to RLP.

test/contracts/OVM/accounts/OVM_ECDSAContractAccount.spec.ts
```diff
@@ -87,48 +79,14 @@ describe('OVM_ECDSAContractAccount', () => {
 
   describe('fallback()', () => {
     it(`should successfully execute an EIP155Transaction`, async () => {
-      const message = serializeNativeTransaction(DEFAULT_EIP155_TX)
-      const sig = await signNativeTransaction(wallet, DEFAULT_EIP155_TX)
+      const transaction = DEFAULT_EIP155_TX
+      const encodedTransaction = await wallet.signTransaction(transaction)
```
smartcontracts, line 84: These tests get way simpler.

test/data/json/libraries/codec/Lib_EIP155Tx.test.json
```diff
@@ -0,0 +1,248 @@
+{
```
kfichter, line 1: Using this test file as the source of these tests: https://github.com/ethereumjs/ethereumjs-tx/blob/master/test/ttTransactionTestEip155VitaliksTests.json

kfichter: I didn't test `hash` here, but `sender` implicitly tests this under the hood so it should be fine.

test/data/json/libraries/codec/Lib_EIP155Tx.test.json
```diff
@@ -0,0 +1,725 @@
...
+                        "0x5208",
+                        "0x3535353535353535353535353535353535353535",
+                        "0x00",
+                        "0x",
+                        "0x25",
+                        "0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d",
+                        "0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d",
+                        "0x01",
+                        "0x00",
+                        false
```
kfichter, line 22: Format is:

```ts
[
    "0x00", // nonce
    "0x04a817c800", // gasPrice
    "0x5208", // gasLimit
    "0x3535353535353535353535353535353535353535", // to
    "0x00", // value
    "0x", // data
    "0x25", // v
    "0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d", // r
    "0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d", // s
    "0x01", // chainId
    "0x00", // recoveryParam
    false // isCreate
]
```

test/helpers/test-runner/json-test-runner.ts
```diff
@@ -4,10 +4,16 @@ import { expect } from '../../setup'
 import { ethers } from 'hardhat'
 import { Contract, BigNumber } from 'ethers'
 
-const bigNumberify = (arr) => {
+const bnRegex = /^\d+n$/gm
```
kfichter, line 7: Made this generally more reliable. Treat more things as big numbers so that equality checking is easier.

test/data/json/libraries/codec/Lib_EIP155Tx.test.json
kfichter: Covers tests for `decode`, `encode`, and `sender`.

test/contracts/OVM/accounts/OVM_ECDSAContractAccount.spec.ts
```diff
@@ -216,52 +154,32 @@ describe('OVM_ECDSAContractAccount', () => {
...
       const ovmREVERT: any =
         Mock__OVM_ExecutionManager.smocked.ovmREVERT.calls[0]
       expect(decodeSolidityError(ovmREVERT._data)).to.equal(
-        'Transaction chainId does not match expected OVM chainId.'
+        'OVM_ECDSAContractAccount: Transaction was signed with the wrong chain ID.'
       )
     })
 
     // TEMPORARY: Skip gas checks for minnet.
     it.skip(`should revert on insufficient gas`, async () => {
```
tynes, line 175: Skipped test has been modified. Should double check it still passes

tynes:
Note that this change must be done in a way that is backwards compatible with the current ctc elements. From the point of view of geth, the old contract must be used when syncing the initial set of transactions. This makes me lean towards not changing the contract and adding a new one that is `OVM_SequencerEntrypointv2`. This way geth can use the old sequencer entrypoint to sync and switch out at a certain point.

This means that geth needs the old serialization logic still and we need a way to selectively set `rawTransaction` in the data transport layer based on the how the transaction is serialized

gakonst:
CI Fails at:

```docker
h_l2_1               | DEBUG[03-16|01:02:16.924] Calling Known Contract                   ID=85f757a4 Name=OVM_ExecutionManager Method=run
geth_l2_1               | DEBUG[03-16|01:02:16.924] Calling Known Contract                   ID=85f757a4 Name=OVM_StateManager     Method=isAuthenticated
geth_l2_1               | DEBUG[03-16|01:02:16.924] Calling Known Contract                   ID=85f757a4 Name=OVM_StateManager     Method=hasAccount
geth_l2_1               | DEBUG[03-16|01:02:16.924] Calling Known Contract                   ID=85f757a4 Name=OVM_StateManager     Method=testAndSetAccountLoaded
geth_l2_1               | DEBUG[03-16|01:02:16.924] Calling Known Contract                   ID=85f757a4 Name=OVM_StateManager     Method=getAccountEthAddress
geth_l2_1               | DEBUG[03-16|01:02:16.924] Calling Known Contract                   ID=85f757a4 Name=OVM_SequencerEntrypoint Message="no method with id: 0x00dede53"
geth_l2_1               | DEBUG[03-16|01:02:16.924] Calling Known Contract                   ID=85f757a4 Name=OVM_ExecutionManager    Method=ovmCHAINID
geth_l2_1               | DEBUG[03-16|01:02:16.925] Reached the end of an OVM execution      ID=85f757a4 Return Data=0x Error="ovm execution failed"
geth_l2_1               | INFO [03-16|01:02:16.925] New block                                index=5     l1-timestamp=1615856583 l1-blocknumber=64 tx-hash=0x40483b4c114a3330ce0afb10500ad4ca304c1a327c5fcab06f0b9b19353074de queue-orign=0 type=0 gas=294814 fees=0 elapsed=4.505ms
```

maurelian:
Converting to draft until checks pass.

maurelian:
Is this dead?

PR #303, opened 2021-03-08
tests: fix the enums to match what is batch submitted

tynes:
## Description

This is a requirement for https://github.com/ethereum-optimism/go-ethereum/pull/255
The enums that were batch submitted to L1 were different than what the contracts expected, so geth had to fix them before passing along to the OVM. The above PR removes any logic in geth around modifying the transactions, it just passes the data through

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
+++ b/contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
@@ -9,12 +9,12 @@ import { Lib_SafeExecutionManagerWrapper } from "../../libraries/wrappers/Lib_Sa
 
 /**
  * @title OVM_SequencerEntrypoint
- * @dev The Sequencer Entrypoint is a predeploy which, despite its name, can in fact be called by 
- * any account. It accepts a more efficient compressed calldata format, which it decompresses and 
+ * @dev The Sequencer Entrypoint is a predeploy which, despite its name, can in fact be called by
+ * any account. It accepts a more efficient compressed calldata format, which it decompresses and
  * encodes to the standard EIP155 transaction format.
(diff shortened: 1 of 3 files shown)


contracts/optimistic-ethereum/OVM/precompiles/OVM_SequencerEntrypoint.sol
```diff
@@ -23,7 +23,7 @@ contract OVM_SequencerEntrypoint {
     /*********
      * Enums *
      *********/
-
```
tynes, line 26: These whitespace diffs add noise to code review. We need linter rules that prevent such things from getting into the codebase

tynes:
I'd expect the integration tests to fail

ben-chain (approved):
All the changes here look good to me; it's always been downstream that blocks here, so if you can get this through CI and have the clear upgrade path, I'm comfortable merging.

tynes:
> All the changes here look good to me; it's always been downstream that blocks here, so if you can get this through CI and have the clear upgrade path, I'm comfortable merging.

We will have to run our state diff tests because this will change the state root from the beginning, so checking state roots that match will not work. I pushed https://github.com/ethereum-optimism/contracts/tree/v0.1.4/enums to base these changes off of the contracts that are deployed to production right now

smartcontracts (commented):
This PR will conflict with #300. What's the specific need for this change?

tynes:
We should prioritize #300 altho I am leaning towards the need to leave in the unmodified `OVM_SequencerEntrypoint.sol` and make a `OVM_SequencerEntrypointv2.sol` or something like that so we can have test coverage of the existing transactions.

I also think its not a bad idea to add a `version() -> string` getter to the contracts to make version management easier as contracts are updated on chain, where the string follows semver

PR #307, opened 2021-03-09
Optimization: Conditional Monotonicity Checking

ben-chain:
## Description
Optimization to the optimization

## Contributing Agreement

### DIFF
--- a/test/contracts/OVM/chain/OVM_CanonicalTransactionChain.spec.ts
+++ b/test/contracts/OVM/chain/OVM_CanonicalTransactionChain.spec.ts
@@ -779,7 +779,7 @@ describe('OVM_CanonicalTransactionChain', () => {
               shouldStartAtBatch: 0,
               totalElementsToAppend: 1,
             })
-          ).to.be.revertedWith('Not enough queued transactions to append.')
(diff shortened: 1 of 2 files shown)


test/contracts/OVM/chain/OVM_CanonicalTransactionChain.spec.ts
```diff
@@ -779,7 +779,7 @@ describe('OVM_CanonicalTransactionChain', () => {
               shouldStartAtBatch: 0,
               totalElementsToAppend: 1,
             })
-          ).to.be.revertedWith('Not enough queued transactions to append.')
+          ).to.be.revertedWith('Index out of bounds.')
```
tynes, line 782: Is this actually the revert message we want? It seems to based on the implementation of the underlying data store

ben-chain: I talked through it with Kelvin; it's because the test case is now failing sooner with this optimization.  I'm gonna merge this in; @smartcontracts if you want to fix this, `&&` on the old and new condition from line `1033` of the CTC would probably fix.

PR #308, opened 2021-03-09
Eliminate EM.safeCREATE logic

ben-chain:
## Description
Simplifies creation logic to not have the EM call itself, and adds a `bytes memory` return field to `ovmCREATE/2` for grabbing revert messages.  This will allow for cleaner error passing in geth etc--yay!

I looked through the compiler to confirm this should be fine with existing contracts, as well as running synthetix integration test suite against it.  This fix can enable simplification of the "safe" bytecode operations, but it is backwards-compatible with the existing strings, so no need for now. 

## Contributing Agreement


contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1027,7 +951,11 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         // behavior can be controlled. In particular, we enforce that flags are passed through
         // revert data as to retrieve execution metadata that would normally be reverted out of
         // existence.
-        (bool success, bytes memory returndata) = _target.call{gas: _gasLimit}(_data);
+
+        (bool success, bytes memory returndata) =
+            _isCreate
+            ? _handleExternalCreate(_gasLimit, _data, _contract)
```
ben-chain, line 957: The basic idea is that `_handleExternalCreate` now wraps the creation result into a consistent `bool, data` return type which will have either `true, 0x` or `false, revert_data`.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1067,7 +995,10 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
 
             // INTENTIONAL_REVERT needs to pass up the user-provided return data encoded into the
             // flag, *not* the full encoded flag. All other revert types return no data.
-            if (flag == RevertFlag.INTENTIONAL_REVERT) {
+            if (
+                flag == RevertFlag.INTENTIONAL_REVERT
+                || _isCreate
```
ben-chain, line 1000: For consistency, the `_handleExternalCreate` function will return an `_encodeRevertData` as if the subcall reverted, even in the case of `CREATE_COLLISION` or `UNSAFE_BYTECODE`.

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
```diff
@@ -113,14 +113,16 @@ contract OVM_ECDSAContractAccount is iOVM_ECDSAContractAccount {
...
-            address created = Lib_SafeExecutionManagerWrapper.safeCREATE(
+            (address created, bytes memory revertData) = Lib_SafeExecutionManagerWrapper.safeCREATE(
                 decodedTx.gasLimit,
                 decodedTx.data
             );
 
             // EVM doesn't tell us whether a contract creation failed, even if it reverted during
             // initialization. Always return `true` for our success value here.
-            return (true, abi.encode(created));
+            return ( created == address(0) ) ?
```
smartcontracts, line 123: I think better to make this an if statement. The ternary syntax is a bit confusing here with the tuple return values.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1021,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+            );
+        }
+
+        // Check the creation bytecode against the OVM_SafetyChecker.
+        if (ovmSafetyChecker.isBytecodeSafe(_creationCode) == false) {
+            return (
+                false,
+                _encodeRevertData(
+                    RevertFlag.UNSAFE_BYTECODE,
+                    Lib_ErrorUtils.encodeRevertString("Contract creation code contains unsafe opcodes.  Did you use the right compiler or pass an unsafe constructor argument?")
```
ben-chain, line 1063: I assumed doing this with the `error(string)` hash would be the way to go here.  Lmk if you think it should be something else @smartcontracts

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1022,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+        returns(
+            bool,
+            bytes memory
+        )
+    {
+        // Check that there is not already code at this address.
+        if (_hasEmptyAccount(_address) == false) {
+            return (
+                false,
+                _encodeRevertData(
```
smartcontracts, line 1050: This differs from the EVM's behavior. Create collision uses all provided gas and returns no data: https://github.com/ethereumjs/ethereumjs-monorepo/blob/53150c159b559b31b6b7c321d784dbb11b082fb9/packages/vm/lib/evm/evm.ts#L270-L278

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1022,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+            );
+        }
+
+        // Check the creation bytecode against the OVM_SafetyChecker.
+        if (ovmSafetyChecker.isBytecodeSafe(_creationCode) == false) {
+            return (
+                false,
+                _encodeRevertData(
+                    RevertFlag.UNSAFE_BYTECODE,
+                    Lib_ErrorUtils.encodeRevertString("Contract creation code contains unsafe opcodes.  Did you use the right compiler or pass an unsafe constructor argument?")
```
smartcontracts, line 1063: Classic @ben-chain with the double space after each period. boo!

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1022,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+
+        // Contract creation didn't need to be reverted and the bytecode is safe. We finish up by
+        // associating the desired address with the newly created contract's code hash and address.
+        _commitPendingAccount(
+            _address,
+            ethAddress,
+            Lib_EthUtils.getCodeHash(ethAddress)
+        );
+
+        // Succressfuly deployments will not give acces to returndata, in both the EVM and the OVM.
```
smartcontracts, line 1112:
```suggestion
        // Successful deployments will not give access to returndata, in both the EVM and the OVM.
```

smartcontracts: Are we sure this is the case? Have we tested this? I don't think it particularly matters (not a consensus issue & I doubt anyone is actually writing contracts that use `RETURNDATACOPY` for successful contract creations.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -363,7 +364,8 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
     /**
      * @notice Overrides CREATE.
      * @param _bytecode Code to be used to CREATE a new contract.
-     * @return _contract Address of the created contract.
+     * @return Address of the created contract.
+     * @return Revert data, iff the creation threw an exception.
```
smartcontracts, line 368: Not sure if `iff` is a typo here or shorthand for "if and only if"

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -399,7 +402,8 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
      * @notice Overrides CREATE2.
      * @param _bytecode Code to be used to CREATE2 a new contract.
      * @param _salt Value used to determine the contract's address.
-     * @return _contract Address of the created contract.
+     * @return Address of the created contract.
+     * @return Revert data, iff the creation threw an exception.
```
smartcontracts, line 406: Same as above.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -904,15 +825,17 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
      * Creates a new contract and associates it with some contract address.
      * @param _contractAddress Address to associate the created contract with.
      * @param _bytecode Bytecode to be used to create the contract.
-     * @return _created Final OVM contract address.
+     * @return Final OVM contract address.
+     * @return Revertdata, iff the creation threw an exception.
```
smartcontracts, line 829:
```suggestion
     * @return Revert data, iff the creation threw an exception.
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -924,25 +847,21 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         nextMessageContext.ovmCALLER = messageContext.ovmADDRESS;
         nextMessageContext.ovmADDRESS = _contractAddress;
 
-        // Run `safeCREATE` in a new EVM message so that our changes can be reflected even if
-        // `safeCREATE` reverts.
-        (bool _success, ) = _handleExternalInteraction(
+        // _handleExternalMessage deals with common checks between call-type and create-type messages.
+        // the _isCreate bool then triggers some create-type-only logic.
```
smartcontracts, line 851: Comment could use some cleanup

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -981,33 +900,38 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-     * @return _returndata Data returned by the call.
+     * Handles all interactions which involve the execution manager calling out to untrusted code (both calls and creates).
+     * Ensures that OVM-related measures are enforced, including L2 gas refunds, nuisance gas, and flagged reversions.
+     * 
+     * @param _nextMessageContext Message context to be used for the external message.
+     * @param _gasLimit Amount of gas to be passed into this message.
+     * @param _contract OVM address being called or deployed to
+     * @param _data Data for the message (either calldata or creation code)
+     * @param _isCreate Whether this is a create-type message.
+     * @return _success Whether or not the message (either a call or deployment) succeeded.
```
smartcontracts, line 921:
```suggestion
     * @return Whether or not the message (either a call or deployment) succeeded.
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -981,33 +900,38 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+     * @return _returndata Data returned by the message.
```
smartcontracts, line 922:
```suggestion
     * @return Data returned by the message.
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -981,33 +900,38 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-        bytes memory _data
+        address _contract,
+        bytes memory _data,
+        bool _isCreate
     )
         internal
-        returns (
-            bool _success,
-            bytes memory _returndata
+        returns(
```
smartcontracts, line 932:
```suggestion
        returns (
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1021,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
 
+    /**
+     * Handles the creation-specific safety measures required for OVM contract deployment.
+     * This function sanitizes the return types for creation messages to match calls (bool, bytes).
+     * This allows for consistent handling of both types of messages in _handleExternalMessage().
+     * 
+     * @param _gasLimit Amount of gas to be passed into this creation.
+     * @param _creationCode Code to pass into CREATE for deployment.
+     * @param _address OVM address being deployed to.
+     * @return _success Whether or not the call succeeded.
```
smartcontracts, line 1032:
```suggestion
     * @return Whether or not the call succeeded.
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1021,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+     * @return _returndata If creation fails: revert data.  Otherwise: empty.
```
smartcontracts, line 1033:
```suggestion
     * @return If creation fails: revert data. Otherwise: empty.
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: Yes, I have tested it a couple times.  This is why I didn't think the approach would work at all.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: Gonna leave it; OOG may cause worse devex due to nuisance gas usage.

contracts/optimistic-ethereum/libraries/utils/Lib_ErrorUtils.sol
```diff
@@ -0,0 +1,34 @@
...
+ */
+library Lib_ErrorUtils {
+
+    /**********************
+     * Internal Functions *
+     **********************/
+
+    /**
+     * Encodes an error string into raw solidity-style revert data.
+     * (i.e. ascii bytes, prefixed with keccak("error(string))")
```
maurelian, line 16: should be `bytes4(keccak("error(string))"))` no?

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -770,90 +775,6 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-        // left over since contract calls can only be passed 63/64ths of total gas,  so we need to
-        // explicitly handle this case here.
-        if (ethAddress == address(0)) {
-            _revertWithFlag(RevertFlag.CREATE_EXCEPTION);
-        }
-
-        // Here we pull out the revert flag that would've been set during creation code. Now that
-        // we're out of creation code again, we can just revert normally while passing the flag
-        // through the revert data.
-        if (messageRecord.revertFlag != RevertFlag.DID_NOT_REVERT) {
```
maurelian, line 836: This flag is no longer used, anywhere in this file. 

It should probably _at least_ be reset in the `_resetContext()`. But maybe there are also implications for that in the new `_handleExternalCreate()`

maurelian (changes requested):
One important question related to resetting the revert flag

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
```diff
@@ -113,14 +113,17 @@ contract OVM_ECDSAContractAccount is iOVM_ECDSAContractAccount {
...
-            address created = Lib_SafeExecutionManagerWrapper.safeCREATE(
+            (address created, bytes memory revertData) = Lib_SafeExecutionManagerWrapper.safeCREATE(
                 decodedTx.gasLimit,
                 decodedTx.data
             );
 
-            // EVM doesn't tell us whether a contract creation failed, even if it reverted during
-            // initialization. Always return `true` for our success value here.
-            return (true, abi.encode(created));
+            // Return true if the contract creation succeeded, false w/ revertData otherwise.
```
smartcontracts, line 121: Could we add this link as a comment somewhere? https://github.com/ethereum/go-ethereum/blob/aae7660410f0ef90279e14afaaf2f429fdc2a186/core/vm/instructions.go#L600-L604

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
```diff
@@ -113,14 +113,17 @@ contract OVM_ECDSAContractAccount is iOVM_ECDSAContractAccount {
...
+            if (created != address(0)) {
+                return (true, abi.encode(created));
```
smartcontracts, line 123: We don't technically need to return this address here, can leave it if we want.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
smartcontracts: Yeah we should be able to remove this flag now.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
smartcontracts: Would you mind leaving a comment that explains this for future readers? ty ser

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1021,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+
+        // Again simply checking that the deployed code is safe too. Contracts can generate
+        // arbitrary deployment code, so there's no easy way to analyze this beforehand.
+        bytes memory deployedCode = Lib_EthUtils.getCode(ethAddress);
+        if (ovmSafetyChecker.isBytecodeSafe(deployedCode) == false) {
+            return (
+                false,
+                _encodeRevertData(
+                    RevertFlag.UNSAFE_BYTECODE,
+                    Lib_ErrorUtils.encodeRevertString("Constrcutor attempted to deploy unsafe opcodes.")
```
smartcontracts, line 1099:
```suggestion
                    Lib_ErrorUtils.encodeRevertString("Constructor attempted to deploy unsafe bytecode.")
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1090,6 +1021,97 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+
+        // Contract creation didn't need to be reverted and the bytecode is safe. We finish up by
+        // associating the desired address with the newly created contract's code hash and address.
+        _commitPendingAccount(
+            _address,
+            ethAddress,
+            Lib_EthUtils.getCodeHash(ethAddress)
+        );
+
+        // Succressfuly deployments will not give acces to returndata, in both the EVM and the OVM.
```
smartcontracts, line 1112:
```suggestion
        // Successful deployments will not give access to returndata, in both the EVM and the OVM.
```

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1835,11 +1836,24 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-        return _transaction.entrypoint.call{gas: _transaction.gasLimit}(_transaction.data);
+        messageContext.ovmADDRESS = _from;
+
+        bool isCreate = _transaction.entrypoint == address(0);
+        if (isCreate) {
+            (address created, bytes memory revertData) = ovmCREATE(_transaction.data);
+            if (created == address(0)) {
+                return (false, revertData);
+            } else {
+                return (true, Lib_EthUtils.getCode(created));
```
smartcontracts, line 1849: Hmmm is `getCode` necessary here?

contracts/optimistic-ethereum/libraries/utils/Lib_ErrorUtils.sol
```diff
@@ -0,0 +1,34 @@
...
+library Lib_ErrorUtils {
+
+    /**********************
+     * Internal Functions *
+     **********************/
+
+    /**
+     * Encodes an error string into raw solidity-style revert data.
+     * (i.e. ascii bytes, prefixed with bytes4(keccak("error(string))"))
+     * @param _reason Reason for the reversion.
```
smartcontracts, line 17: Request to include this link somewhere: https://docs.soliditylang.org/en/v0.8.2/control-structures.html?highlight=Error(string)#panic-via-assert-and-error-via-require

smartcontracts (commented):
Asking for a few minor changes. Otherwise LGTM.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
maurelian: I haven't looked for a reference regarding returndata from a `CREATE`, but my intuition about why you can't access it is because the RETURN opcode is used to get the actual runtime bytecode which is deployed. If one wanted _that_ info, they could use `EXTCODECOPY`.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
maurelian: I'm not sure of that. We at least need a flag corresponding to the default context prior to `initializeContext` and after `resetContext`.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1806,7 +1808,6 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         messageContext.isStatic = false;
 
         messageRecord.nuisanceGasLeft = 0;
-        messageRecord.revertFlag = RevertFlag.DID_NOT_REVERT;
```
maurelian, line 1809: I think we still need this.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: With this change, the value luckily does is not stored anymore (see change to `iOVM_EXecutionManager.messageRecord`), so I think it can genuinely be removed now.

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
ben-chain: put it in the return line of `ovmCREATE`

contracts/optimistic-ethereum/OVM/accounts/OVM_ECDSAContractAccount.sol
ben-chain: Yeah, I lean towards leaving it.  You can see how it's actually done even differently in `eth_call`, see the code lookup in the `simulateMessage` case.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: It was the easiest way I saw to do it.  The return type is correct if that's what you're getting at -- the `eth_call` endpoint for `to = undefined` will return the deployed bytecode if successful, differing from a standard evmCREATE message.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: What would be the condition in which it's used?  This variable previously existed solely for the purposes of being set in `safeCREATE`, so without that logic, I'm pretty sure we're safe to remove it.

maurelian (approved):

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
smartcontracts: Oh interesting. I wasn't aware of that. Sounds good.

smartcontracts: Actually: could we put a comment here explaining just that? Even copy/paste what you wrote here would be fine.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1426,7 +1451,6 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         // Out of gas and create exceptions will fundamentally return no data, so simulating it shouldn't either.
         if (
             _flag == RevertFlag.OUT_OF_GAS
-            || _flag == RevertFlag.CREATE_EXCEPTION
```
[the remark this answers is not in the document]
ben-chain: This flag was originally the `RevertFlag` used by the `safeCREATE` function to indicate that contract creation had thrown an exception.  With this PR, that code path is replaced by `_handleContractCreation` returning `false, revertData`.  That the creation threw an exception is now implicit by that return value, and doesn't need its own flag.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1089,6 +1021,99 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+        }
+
+        // We always need to initialize the contract with the default account values.
+        _initPendingAccount(_address);
+
+        // Actually execute the EVM create message,
+        address ethAddress = Lib_EthUtils.createContract(_creationCode);
+        
+        if (ethAddress == address(0)) {
+            // If the creation fails, the EVM lets us grab its revert data.  This may contain a revert flag
```
smartcontracts, line 1077:
```suggestion
            // If the creation fails, the EVM lets us grab its revert data. This may contain a revert flag
```

smartcontracts (approved):

ben-chain:
Alright, just locally ran the final build through synthetix integration tests as well.  Time to ship!

PR #312, opened 2021-03-10
fix: Make ovmSETNONCE monotonic

smartcontracts:
## Description
Need review from @ben-chain and will need to add tests for this. Need to make sure this doesn't introduce any downstream effects. Proto pointed out (rightfully) that a *lot* of client-side tooling expects nonces to increase and not decrease. We may perhaps also only want to let people bump their nonce by 1 at a time.

## Contributing Agreement


protolambda:
Not sure if this improves the situation. A call to set-nonce that silently does nothing seems no better than a call that allows you do to something unexpected explicitly.

smartcontracts:
> Not sure if this improves the situation. A call to set-nonce that silently does nothing seems no better than a call that allows you do to something unexpected explicitly.

Moving over to an `ovmINCREMENTNONCE` instead. Will bump nonce by 1 but will not do anything if nonce hits the uint64 max. What do you think?

protolambda:
I'm not sure about changing the function altogether either. Maybe we can discuss on next call how the whole nonce system can/should behave.

smartcontracts:
Sounds good to me. I'll make sure @ben-chain is around for the next call too, he'll have some good input on this subject.

maurelian:
Based on the conversation above I've converted this to a draft PR. Feel free to reverse that if/when this is ready for review.

smartcontracts:
Would be great to get a final determination on this ASAP @ben-chain

tynes:
This feels a lot safer to me personally. If allowing the nonce to be managed arbitrarily, what exactly is it useful for? If its not obviously secure, we should cut it for now

smartcontracts:
> This feels a lot safer to me personally. If allowing the nonce to be managed arbitrarily, what exactly is it useful for? If its not obviously secure, we should cut it for now

I completely agree.

smartcontracts:
Pinging @ben-chain again.

ben-chain (approved):
I think this is the right way to go too.  In principle you could have a `revertFlag` which forces call frames which attempt to overflow the nonce to throw an exception, but it's unclear to me if that improves the situation or e.g. introduces a way for users' fee payment to be reverted. I think given the `uint64` it's such an edge case anyway that we can proceed here.

PR #314, opened 2021-03-11
Reset context on return

maurelian:
## Description

Fixes a bug that left the EM context initialized after a call. 

Fixes: https://github.com/ethereum-optimism/roadmap/issues/781

I looked for other assembly level `return` opcodes, and did not find any.


ben-chain:
This logic is being removed in https://github.com/ethereum-optimism/contracts/pull/308, so I'm gonna close this out.

PR #315, opened 2021-03-11
Fix simulateMessage for eth_calls

ben-chain:
## Description
This allows deployed code to be correctly returned by geth in L2 during `eth_call`.

## Contributing Agreement


gakonst (approved):
LGTM. Awesome work narrowing it down. Can we also have a test, since there's [no tests](https://github.com/search?q=org%3Aethereum-optimism+simulateMessage&type=code) for this function?

maurelian (approved):
LGTM. Agree a test would be helpful, but I won't block this for it.

Issue #309, opened 2021-03-10
Consider resetting xDomainMessageSender after relaying a message 


cwhinfrey:
**Is your feature request related to a problem? Please describe.**
`xDomainMessageSender` will continue to return the sender of the last message until the next message is relayed.

This shouldn't be an issue for the most part since `relayMessage` has a reentrancy guard and every contract consuming `xDomainMessageSender` should also be checking that `msg.sender` is the Optimism messenger. That being said, it could lead to some weird behavior and is easy to fix.

**Describe the solution you'd like**
Consider either zeroing out xDomainMessageSender after the calls [here](https://github.com/ethereum-optimism/contracts/blob/master/contracts/optimistic-ethereum/OVM/bridge/messaging/OVM_L1CrossDomainMessenger.sol#L117) and [here](https://github.com/ethereum-optimism/contracts/blob/master/contracts/optimistic-ethereum/OVM/bridge/messaging/OVM_L2CrossDomainMessenger.sol#L79) or setting is to an address like `0x000000000000000000000000000000000000dead`.  Either one should result in some [gas savings](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2200.md) since the transaction will always reset the storage slot to the same value that it was at the beginning of the transaction. Each option will have slightly different [gas implications](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/ReentrancyGuard.sol#L28-L32) (I'd recommend setting to the dead address).

Basically, it should improve security and save a bit of gas.

**Describe alternatives you've considered**
Leave as is

PR #316, opened 2021-03-11
perf: reset xDomainMessageSender to default value after external call

gakonst:
Fixes https://github.com/ethereum-optimism/contracts/issues/309


contracts/optimistic-ethereum/OVM/bridge/messaging/Abs_BaseCrossDomainMessenger.sol
```diff
@@ -33,7 +41,9 @@ abstract contract Abs_BaseCrossDomainMessenger is iAbs_BaseCrossDomainMessenger,
      * Public Functions *
      ********************/
 
-    constructor() Lib_ReentrancyGuard() {}
+    constructor() Lib_ReentrancyGuard() internal {
+        xDomainMessageSender = DEFAULT_XDOMAIN_SENDER;
```
smartcontracts, line 45: Possible devex improvement here would be to make `xDomainMessageSender()` revert if the sender is this default value. Not sure how much it matters, but I feel like contracts shouldn't be able to use `xDomainMessageSender()` if this contract isn't trying to send a message during that tx.

contracts/optimistic-ethereum/OVM/bridge/messaging/Abs_BaseCrossDomainMessenger.sol
```diff
@@ -18,6 +18,14 @@ import { Lib_ReentrancyGuard } from "../../../libraries/utils/Lib_ReentrancyGuar
  * Runtime target: defined by child contract
  */
 abstract contract Abs_BaseCrossDomainMessenger is iAbs_BaseCrossDomainMessenger, Lib_ReentrancyGuard {
+    /**********************
+     *      Constants     *
+     **********************/
```
smartcontracts, line 23: huff puff

```
/*************
 * Constants *
 *************/
```

contracts/optimistic-ethereum/OVM/bridge/messaging/OVM_L1CrossDomainMessenger.sol
```diff
@@ -49,6 +49,7 @@ contract OVM_L1CrossDomainMessenger is iOVM_L1CrossDomainMessenger, Abs_BaseCros
     {
         require(address(libAddressManager) == address(0), "L1CrossDomainMessenger already intialized.");
         libAddressManager = Lib_AddressManager(_libAddressManager);
+        xDomainMessageSender = DEFAULT_XDOMAIN_SENDER;
```
maurelian, line 52: I think this should be a no-op given it happens in the constructor, but I think it'll only cost 200 gas here.

contracts/optimistic-ethereum/OVM/bridge/messaging/OVM_L1CrossDomainMessenger.sol
```diff
@@ -49,6 +49,7 @@ contract OVM_L1CrossDomainMessenger is iOVM_L1CrossDomainMessenger, Abs_BaseCros
     {
```
maurelian, line 49: Funny that we have an `initialize()` function, but no `isInitialized()` checks on the other functions.

contracts/optimistic-ethereum/OVM/bridge/messaging/OVM_L2CrossDomainMessenger.sol
```diff
@@ -35,7 +35,9 @@ contract OVM_L2CrossDomainMessenger is iOVM_L2CrossDomainMessenger, Abs_BaseCros
         address _libAddressManager
     )
         Lib_AddressResolver(_libAddressManager)
-    {}
+    {
+        xDomainMessageSender = DEFAULT_XDOMAIN_SENDER;
```
maurelian, line 39: Similar comment, I think this is a no-op.

maurelian (dismissed):

contracts/optimistic-ethereum/OVM/bridge/messaging/OVM_L1CrossDomainMessenger.sol
gakonst: This doesn't work with the L1 XDomain Messenger. I tried it, not sure why your constructor is not called.

contracts/optimistic-ethereum/OVM/bridge/messaging/Abs_BaseCrossDomainMessenger.sol
gakonst: https://github.com/ethereum-optimism/contracts/pull/320/commits/83cff29d364f3f85d496f1aa3ca64d669903de0d what do you think about this? It requires that we remove the tests, since the test happens before/after the call and that means it always reverts.

contracts/optimistic-ethereum/OVM/bridge/messaging/OVM_L1CrossDomainMessenger.sol
maurelian: The initialize function is needed on at least one of them, because when you deploy the first one, you can't yet know the address of the other one.

contracts/optimistic-ethereum/OVM/bridge/messaging/Abs_BaseCrossDomainMessenger.sol
```diff
@@ -27,13 +35,18 @@ abstract contract Abs_BaseCrossDomainMessenger is iAbs_BaseCrossDomainMessenger,
     mapping (bytes32 => bool) public successfulMessages;
     mapping (bytes32 => bool) public sentMessages;
     uint256 public messageNonce;
-    address override public xDomainMessageSender;
+    address internal xDomainMsgSender = DEFAULT_XDOMAIN_SENDER;
```
maurelian, line 38: I assume the name change is necessary to avoid a shadowed variable warning between this variable and the new getter function?

maurelian (approved):

PR #317, opened 2021-03-11
Add a basic .editorconfig file

maurelian:
## Description

Based on Augur's `.editorconfig` (https://github.com/AugurProject/augur/blob/dev/.editorconfig). 

This will pretty much eliminate any issues with extra whitespace creeping in at EOL, resulting in 'dirty' commits. Though it might also mean that commits are noisier for a while, as the extra whitespace will be removed anytime a file is opened and saved again.

Many editors will inherit settings from this file, but VSCode requires an [extension](VSCode requires an extension to support it.) to support it. Learn more: https://editorconfig.org/.


maurelian:
Keeping as draft so I can consider the implications of 

> it might also mean that commits are noisier for a while, as the extra whitespace will be removed anytime a file is opened and saved again.

smartcontracts:
> it might also mean that commits are noisier for a while, as the extra whitespace will be removed anytime a file is opened and saved again.

I think this is fine. Not a huge deal. Could circumvent this by going through each file and fixing extra whitespace in a single PR if we want.

smartcontracts:
@maurelian is this ready for review?

maurelian:
Yes, I think we should do this.

smartcontracts (approved):

PR #320, opened 2021-03-12
Fix/revert if xdomain unset

gakonst:
> Possible devex improvement here would be to make xDomainMessageSender() revert if the sender is this default value. Not sure how much it matters, but I feel like contracts shouldn't be able to use xDomainMessageSender() if this contract isn't trying to send a message during that tx.


gakonst:
Opened by accident

PR #322, opened 2021-03-12
fix: Clear Compiler Warnings

### LINKED ISSUE #255: Clear up compiler warnings
Compiling our contracts currently returns a large number of compiler warnings. 

Fixing them should be a fairly simple task, but will require some time to go through everything. 

**Recommended approach:**

IMO the solc binary give much easier to read colored output compared to hardhat, so install `solc` 
1. Install `solc` verson 0.7.6 following [these instructions](https://docs.soliditylang.org/en/v0.7.6/installing-solidity.html) (any of the methods described on that page _other than_ `npm i -g solc` because `solcjs != solc`!). 
2. `cd` into the root of this repo. 
3. Run the compiler on the full system using: `solc --allow-paths . contracts/optimistic-ethereum/**/*.sol` 
4. Work through the warnings systematically. You might choose to focus on one file at a time, or one type of warning at a time, whichever works for you. 

I would suggest submitting a draft PR for early review after clearing up the first dozen or so warnings, just to be sure we're on track.

vcshih:
## Description
Clears compiler warnings when compiling contracts. Each warning type is addressed within a single commit in this PR.

## Questions
- I wasn't sure if I addressed `Warning: Unnamed return variable can remain unassigned.` correctly. Specifically, I took out a return for one of the functions because it was not being used.
- For `unused local variable warning`, I took out a couple of local variables for calculating `gasUsed` which was not being used in the function. Is that okay?

### Fixes
- Fixes #255

## Contributing Agreement

### DIFF
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -470,7 +470,6 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
                 curContext,
                 nextContext,
                 nextQueueIndex,
-                queueLength,
                 queueRef
             );
 
@@ -1008,14 +1007,12 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
      * @param _prevContext The previously validated batch context.
      * @param _nextContext The batch context to validate with this call.
      * @param _nextQueueIndex Index of the next queue element to process for the _nextContext's subsequentQueueElements.
-     * param _queueLength The length of the queue at the start of the batchAppend call.
      * @param _queueRef The storage container for the queue.
      */
     function _validateNextBatchContext(
         BatchContext memory _prevContext,
         BatchContext memory _nextContext,
         uint40 _nextQueueIndex,
-        uint40, // _queueLength,
         iOVM_ChainStorageContainer _queueRef
     )
         internal
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -341,6 +341,7 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
     )
         override
         public
+        pure
     {
         // TEMPORARY: Disable `appendQueueBatch` for minnet
         revert("appendQueueBatch is currently disabled.");
--- a/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
+++ b/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
@@ -334,10 +334,10 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
 
     /**
      * Appends a given number of queued transactions as a single batch.
-     * @param _numQueuedTransactions Number of transactions to append.
+     * param _numQueuedTransactions Number of transactions to append.
      */
     function appendQueueBatch(
-        uint256 _numQueuedTransactions
+        uint256 // _numQueuedTransactions
     )
         override
         public
@@ -1007,14 +1007,14 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
      * @param _prevContext The previously validated batch context.
      * @param _nextContext The batch context to validate with this call.
      * @param _nextQueueIndex Index of the next queue element to process for the _nextContext's subsequentQueueElements.
-     * @param _queueLength The length of the queue at the start of the batchAppend call.
+     * param _queueLength The length of the queue at the start of the batchAppend call.
      * @param _queueRef The storage container for the queue.
      */
     function _validateNextBatchContext(
         BatchContext memory _prevContext,
         BatchContext memory _nextContext,
         uint40 _nextQueueIndex,
-        uint40 _queueLength,
+        uint40, // _queueLength,
         iOVM_ChainStorageContainer _queueRef
     )
         internal
--- a/contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
+++ b/contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
@@ -187,16 +187,12 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
             return;
         }
 
-        // Check gas right before the call to get total gas consumed by OVM transaction.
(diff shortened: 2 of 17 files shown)


contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
```diff
@@ -1007,14 +1008,14 @@ contract OVM_CanonicalTransactionChain is iOVM_CanonicalTransactionChain, Lib_Ad
...
      * @param _nextQueueIndex Index of the next queue element to process for the _nextContext's subsequentQueueElements.
-     * @param _queueLength The length of the queue at the start of the batchAppend call.
+     * param _queueLength The length of the queue at the start of the batchAppend call.
      * @param _queueRef The storage container for the queue.
      */
     function _validateNextBatchContext(
         BatchContext memory _prevContext,
         BatchContext memory _nextContext,
         uint40 _nextQueueIndex,
-        uint40 _queueLength,
```
smartcontracts, line 1017: Seems like this variable could just be removed entirely from the params here. That might be the better way to clear out this compiler warning.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -187,16 +187,12 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
             return;
         }
 
-        // Check gas right before the call to get total gas consumed by OVM transaction.
-        uint256 gasProvided = gasleft();
```
smartcontracts, line 191: I think this should be commented out instead of deleted.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -187,16 +187,12 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-
         // Run the transaction, make sure to meter the gas usage.
         ovmCALL(
             _transaction.gasLimit - gasMeterConfig.minTransactionGasLimit,
             _transaction.entrypoint,
             _transaction.data
         );
-        uint256 gasUsed = gasProvided - gasleft();
```
smartcontracts, line 199: Same as above, commented instead of deleted.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -632,8 +628,6 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         // DELEGATECALL does not change anything about the message context.
         MessageContext memory nextMessageContext = messageContext;
         
-        bool isStaticEntrypoint = false;
```
smartcontracts, line 635: @ben-chain was this recently made obsolete?

contracts/test-helpers/Helper_PrecompileCaller.sol
```diff
@@ -43,10 +43,9 @@ contract Helper_PrecompileCaller is Helper_SimpleProxy {
         bytes memory _data
     )
         public
-        returns (
```
smartcontracts, line 46: Hmm I don't think this needs to be changed.

smartcontracts (commented):
Thank you for this! Mostly LGTM, left a few comments. Not 100% sure about all of the changes, but commented where I was concerned.

contracts/test-helpers/Helper_PrecompileCaller.sol
vcshih: I reverted the address name change, but were you talking about the `return msg.sender`? The compiler gets unhappy when there is a `returns` statement with no return

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: Oooh I recognize this variable!  This was from when we thought nested `ovmCALLs` within an `ovmSTATICCALL` should all collapse up atomically if they attempted a violation.  It's been obsolete for a while, actually; surprised to see it in here still--makes sense to me that it would've been an `unused` error.  Yay code removal! :D

contracts/test-helpers/Helper_PrecompileCaller.sol
```diff
@@ -48,5 +48,6 @@ contract Helper_PrecompileCaller is Helper_SimpleProxy {
         )
     {
         callPrecompile(_precompile, _data);
+        return msg.sender;
```
smartcontracts, line 51: I have to double check this one.

smartcontracts: @ben-chain would be good to get your eyes on this, not sure where/if this is used. But I think `callPrecompile` does a low-level `return` inside of assembly so this return statement won't do anything (I think?)

contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol
ben-chain: Ah yes, confirming that this is vestigial now with https://github.com/ethereum-optimism/contracts/pull/307/ pulling the logic into `_validateFinalBatchContext`.

contracts/optimistic-ethereum/mockOVM/bridge/mockOVM_GenericCrossDomainMessenger.sol
```diff
@@ -40,7 +40,8 @@ contract mockOVM_GenericCrossDomainMessenger {
         public
     {
         xDomainMessageSender = _sender;
-        _target.call{gas: _gasLimit}(_message);
+        (bool success, ) = _target.call{gas: _gasLimit}(_message);
```
ben-chain, line 43: If I'm not mistaken, this is a devex improvement disguised as a compiler fix but we'll allow it 😛

@smartcontracts I think you have the knowledge of this contract; seems fine to me but you know best.

smartcontracts: Yeah this is fine. Contract only gets used inside of our plugins to simulate L1<>L2 comms anyway.

contracts/test-helpers/Helper_PrecompileCaller.sol
ben-chain: Yeah, this return value being added definitely goes unused.  @vcshih if the compiler shuts up with just `return;`, we can do that; otherwise `return address(0); // unused` or something feels like the way to go.

smartcontracts: Agree with `return address(0); // unused: silence compiler` or similar.

vcshih: Makes sense to me :+1: Implemented in the most recent commit

ben-chain (approved):
Heroic stuff, thank you!! 😄. Looks good to ship pending CI.

PR #324, opened 2021-03-13
remove pushTwo and getTwo

K-Ho:
## Description

Current Chain storage containers on Kovan and Mainnet don't have pushTwo or getTwo, and they add a bit of complexity, so let's just remove them for now. The gas savings are minimal according to @smartcontracts, so I think we should be good there.

## Questions
- 
-
-

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement


ben-chain (approved):
LGTM!

PR #327, opened 2021-03-15
refactor: use AddressZero instead of ZERO_ADDRESS

snario:
Uses a more standardized zero address.


maurelian:
`ZERO_ADDRESS` also appears in 20 or so other *.ts files according to my search, so I think we should make this change to all of them at once.

TransmissionsDev:
TIL!

karlfloersch (commented):
Nice change & I do think we should change it in a bunch of places

smartcontracts:
@karlfloersch I made #329 as a "Good First Issue" for people on discord to tackle.

vaibhavchellani:
Will this PR fix #329 or is it up for taking?

PR #332, opened 2021-03-16
Rename OVM precompiles to predeploys

maurelian:
## Description

A less confusing nomenclature is the following:
- The EVM has precompiles. 
- The OVM has predeploys (except for the State Manager, which is a legitimate "precompile" in L2 geth). 

This PR simply changes directory names in the contracts, and variable names in test files _where appropriate_. I have done my best to not make a change in the places where we actually mean precompile. 

One important string not changed is "OVM_DecompressionPrecompileAddress", which I gather is the Address Manager's name for the Sequencer Entrypoint:

https://github.com/ethereum-optimism/contracts-v2/blob/38ee0ef122af4274b65e3764ebfabeee46985863/contracts/optimistic-ethereum/OVM/chain/OVM_CanonicalTransactionChain.sol#L1129


smartcontracts (approved):
Looks good. Shouldn't cause any conflicts elsewhere.

ben-chain (approved):
LGTM, let's merge!  Don't think there are any downstream imports that the future package update will break.

PR #333, opened 2021-03-16
rename OVM_ETH symbol and name

K-Ho:
## Questions
- 
-
-

### Fixes
- Fixes # [Link to Issue]

## Contributing Agreement


smartcontracts (approved):

tynes (approved):

tynes:
I do have a question around what is ETH and what is WETH on Optimism - right now it seems like WETH is ETH and ETH is WETH?

maurelian:
@tynes as currently implemented, all ETH on L2 is actually WETH. I believe @ben-chain's plans are to make it so that you can interact with ETH as if it were both WETH and ETH.

maurelian (approved):
I don't object, so I'll approve, but what is the rationale?

smartcontracts:
> I don't object, so I'll approve, but what is the rationale?

Since the L2 should be backed by L1 security, the ETH on L2 should generally be treated the same as on L1. I also think it's a bit less confusing to say "you have ETH on L2" rather than "you have this different wrapped ETH asset on L2 that can be redeemed at a 1:1 rate for ETH on L1"

maurelian:
@ben-chain please review!

Also, a question: could this have any implications for other components?

ben-chain (approved):
I think this is the right move, approving!

PR #336, opened 2021-03-18
Add a warning about getMerkleRoot

maurelian:
Based on a comment from Zeppelin: 

> Note that `getMerkleRoot` accepts a buffer of memory, and then modifies it in place while calculating the root hash. So if the outer function tried to use the input buffer after they call `getMerkleRoot`, it would be corrupted.

> Tracing it through the code base, it seems possible this might happen with `OVM_StateCommitmentChain.appendStateBatch`, which just takes the `_batch` from the user so you can't control whether they plan to use it again.

TBH,  I don't _love_ adding this long comment everywhere the function gets used, but it seems dangerous enough that it's worthwhile.

Otherwise I'd settle for just having it in the natspec.


ben-chain (approved):
LGTM, thanks for this!

smartcontracts:
Related to #344

PR #342, opened 2021-03-23
feat: Have ExecutionManager pass data up

smartcontracts:
**Description**
Ensures that `OVM_ExecutionManager` will pass data upwards. Also modifies `OVM_ProxyEOA` and `OVM_SequencerEntrypoint` to pass data upwards and out of the EM. 

**Additional context**
Vastly simplifies the process of giving users relevant return data in Geth. Cuts a lot of code out of `evm.go`.

**Metadata**
- Fixes roadmap728


smartcontracts:
Tests might fail, will fix if they do. Want CI to run in the background while I work on some other stuff.

contracts/optimistic-ethereum/OVM/predeploys/OVM_SequencerEntrypoint.sol
```diff
@@ -89,13 +89,23 @@ contract OVM_SequencerEntrypoint {
...
-        Lib_SafeExecutionManagerWrapper.safeCALL(
+        (bool success, bytes memory returndata) = Lib_SafeExecutionManagerWrapper.safeCALL(
             gasleft(),
             target,
             callbytes
         );
+
+        if (success) {
+            assembly {
+                return(add(returndata, 0x20), mload(returndata))
```
maurelian, line 100: Why do we need assembly here? 
Is it possible that the length of `returndata` is less than 32 bytes?

smartcontracts: Can't use `return` in a fallback function without assembly.

maurelian: I see. 
> Is it possible that the length of returndata is less than 32 bytes?
Whoops, I was thinking of storage encoding here.

contracts/optimistic-ethereum/libraries/wrappers/Lib_SafeExecutionManagerWrapper.sol
```diff
@@ -355,10 +372,6 @@ library Lib_SafeExecutionManagerWrapper {
             assembly {
                 revert(add(returndata, 0x20), mload(returndata))
```
maurelian, line 373: I know it's outside of scope, but why does this need assembly?

maurelian (approved):
LGTM.

tynes:
@smartcontracts Can you confirm that this PR is fully backwards compatible?

smartcontracts:
> @smartcontracts Can you confirm that this PR is fully backwards compatible?

I do *not* think this PR is fully backwards compatible. So we're going to have to keep this one open for a little.

PR #343, opened 2021-03-23
fix: Stricter RLP boolean checking

smartcontracts:
**Description**
Minor fix, handles an issue pointed out by proto. Geth is much stricter than `Lib_RLPReader` about checking the validity of boolean values. This PR fixes that discrepancy and explicitly checks that the boolean value is either 0 or 1.

**Metadata**
- Fixes proto#42


gakonst (approved):
LGTM

maurelian (approved):
LGTM

PR #347, opened 2021-03-25
chore: use trimmed down smock/plugins

gakonst:
Makes the repository more lightweight for dev installations. This is blocked on the `smock` release getting fixed.


package.json
```diff
@@ -56,7 +56,7 @@
     "buffer-xor": "^2.0.2",
     "chai": "^4.3.1",
     "copyfiles": "^2.3.0",
-    "ethereum-waffle": "3.0.0",
+    "ethereum-waffle": "^3.3.0",
```
maurelian, line 59: I see that this version was pinned and is now floating. 

@smartcontracts the blame says this was done in the initial commit. I'm curious if there's a reason the version was previously pinned?

smartcontracts: Nope. Probably my mistake.

smartcontracts (approved):

PR #296, opened 2021-03-04
feat: allow appending extraData on withdrawals

gakonst:
## Description

This enables an L2 user to provide additional data, which can be then
used by L1 contracts. The L1 contract would provide the same data, and it
would then be checked against the hash of `relayedMessages` on the L1 Messenger

PR #349, opened 2021-03-25
Add 'from' to `finalizeWithdrawal` in Standard Bridge Interface

krzkaczor:
**Description**
Add `from` field to `iOVM_L1TokenGateway`'s `finalizeWithdrawal`. This adds more context to cross-chain messages and allows to build 3rd party fast withdrawal solutions OUTSIDE of the token bridge.

**Additional context**
FW solution flow possible after this change:
1. Alice calls `withdrawTo` on `L2DepositedToken` with an address of a smartcontract coordinating FW let's call it `Coordinator`.
2. A liqudity provider Bob validates that withdrawal is correct.
3. Bob sends Alice funds on L1 by letting `Coordinator` to send his funds to Alice and record this fact. Alice can enjoy their token on L1.
4. When Alice's withdrawal settles, Bob claims it as he can proof that he already sent Alice funds.

If (3) never happened, after withdrawal settles, Alice proofs that she initiated the original withdraw and she claims tokens. Without the change described in this PR a crosschain message only contains address of a Coordinator SC (`to` field) and hence Alice can not proof that she originated the withdrawal. 

PoC of such solution was implemented by @gakonst in: https://github.com/gakonst/optimistic-fast-withdrawals/blob/feat/fast-withdrawals/contracts/MarketMaker.sol

Alice and Bob can negotiate additional fees in a state channel and they don't have to be part of the cross-chain message.

**Other implementations**
A similar effect can be achieved by adding `extraData` instead of `from` but this could produce multiple similar yet different implementations in the wild.

### DIFF
--- a/contracts/optimistic-ethereum/OVM/bridge/tokens/Abs_L1TokenGateway.sol
+++ b/contracts/optimistic-ethereum/OVM/bridge/tokens/Abs_L1TokenGateway.sol
@@ -64,6 +64,7 @@ abstract contract Abs_L1TokenGateway is iOVM_L1TokenGateway, OVM_CrossDomainEnab
      */
     function _handleFinalizeWithdrawal(
         address, // _to,
+        address, // _from,
         uint256 // _amount
     )
         internal
@@ -188,22 +189,25 @@ abstract contract Abs_L1TokenGateway is iOVM_L1TokenGateway, OVM_CrossDomainEnab
      * This call will fail if the initialized withdrawal from L2 has not been finalized. 
      *
      * @param _to L1 address to credit the withdrawal to
+     * @param _from address that started withdrawal
      * @param _amount Amount of the ERC20 to withdraw
      */
     function finalizeWithdrawal(
         address _to,
+        address _from,
         uint _amount
     )
         external
-        override 
+        override
         onlyFromCrossDomainAccount(l2DepositedToken)
     {
         // Call our withdrawal accounting handler implemented by child contracts.
         _handleFinalizeWithdrawal(
             _to,
+            _from,
             _amount
         );
 
-        emit WithdrawalFinalized(_to, _amount);
+        emit WithdrawalFinalized(_to, _from, _amount);
     }
 }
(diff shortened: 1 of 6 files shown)


TransmissionsDev:
I think ultimately the better solution is to add an arbitrary bytes field for the Coordinator to populate with Alice's address and any other data she wants. There's a lot of situations we can (and I already have) run into where you need more info than just the address of the origin.

krzkaczor:
Crap, I didn't realize that @gakonst already made a PR suggesting something similar 😅 

Still, I don't understand how sending arbitrary bytes is better than a standardized interface. Imagine FW being implemented as a feature in wallets. With my proposal, it's dead simple as all tokens and all `Coordinators` behave the same way. But if `extraData` can be anything this means that different `Coordinators` will expect different data payloads. 

In my proposal, details of the deal (extraData?) can be discussed it a state channel so `Coordinator` just verifies Alice's and Bob's signatures.

krzkaczor:
Any idea why @gakonst's original PR was closed?

gakonst:
> Still, I don't understand how sending arbitrary bytes is better than a standardized interface. Imagine FW being implemented as a feature in wallets

Arbitrary bytes allow you to specify more "complex" messages, e.g. require that your transaction can only pay up to some fee, or can be picked up by a whitelist of relays etc.

maurelian:
Thank you for the PR @krzkaczor!

What's the current proposal though? Between the conversation here and on #296, I'm not sure which direction folks are leaning towards. 

My preference would actually be to have an issue opened describing: 
1. The problem
2. The possible solutions and their relative advantages. 

That way we're a bit closer to working from a spec, and PRs can be reviewed for correctness alone without having to debate the merits of the direction.

krzkaczor:
@maurelian yes, makes sense. I will close it for now and create an issue dumping all that I know and kicking off discussion in public.

krzkaczor:
@maurelian here's the issue: https://github.com/ethereum-optimism/contracts/issues/355

PR #350, opened 2021-03-26
Refactor repeatable hh-deploy code to helper function

gakonst:
As title, should make the https://github.com/ethereum-optimism/contracts/pull/249 more easy to review. 

This can be abstracted a bit more on the proxies. 

Q: Why are we using a mock ovm bond manager?

### DIFF
--- a/deploy/001-OVM_ChainStorageContainer_ctc_batches.deploy.ts
+++ b/deploy/001-OVM_ChainStorageContainer_ctc_batches.deploy.ts
@@ -2,35 +2,16 @@
 import { DeployFunction } from 'hardhat-deploy/dist/types'
 
 /* Imports: Internal */
-import { getDeployedContract } from '../src/hardhat-deploy-ethers'
+import { deploy } from '../src/hardhat-deploy-ethers'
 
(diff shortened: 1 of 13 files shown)


smartcontracts:
FWIW I wanted to make the deploy scripts straightforward and legible -- duplicating the code was a deliberate design choice here. I could be open to changing my mind if you feel strongly about it and have an argument.

tynes (commented):
I think that deduplication in this case is a good thing, if the concern is that functionality is hidden, then I think we just rename `deploy` to `deployAndRegister` or something like that

tynes:
> Q: Why are we using a mock ovm bond manager?

It is because the mock bond manager is what is currently deployed on mainnet and we need the deployment infra to be set up to exactly what we have now so that we can use it for upgrades

smartcontracts:
`deploy` is definitely a confusing name here because of the `deploy` function exported by `hardhat-deploy`. So definitely needs a rename. `deployAndResister` seems fine.

tynes:
@smartcontracts want to merge this and then just add that fix yourself?

PR #352, opened 2021-03-26
Make the libAddressManager variable public

smartcontracts:
**Description**
The `libAddressManager` variable is currently internal and isn't easily accessible from the outside. Ideally this variable should be public so we have more visibility into the state of our contracts. Particularly important for doing post-deployment verification for proxies (which are initialized outside of the constructor).


ben-chain (approved):
👍 Good catch; makes sense!

PR #353, opened 2021-03-29
Fix shortcut for single element

Dominator008:

### DIFF
--- a/contracts/optimistic-ethereum/libraries/utils/Lib_MerkleTree.sol
+++ b/contracts/optimistic-ethereum/libraries/utils/Lib_MerkleTree.sol
@@ -34,7 +34,7 @@ library Lib_MerkleTree {
             "Lib_MerkleTree: Must provide at least one leaf hash."
         );
 
-        if (_elements.length == 0) {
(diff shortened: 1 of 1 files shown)


contracts/optimistic-ethereum/libraries/utils/Lib_MerkleTree.sol
```diff
@@ -34,7 +34,7 @@ library Lib_MerkleTree {
             "Lib_MerkleTree: Must provide at least one leaf hash."
         );
 
-        if (_elements.length == 0) {
+        if (_elements.length == 1) {
             return _elements[0];
```
maurelian, line 38: Good catch. It looks like this line is not even reachable prior to the change.

maurelian (approved):
This looks like a good catch to me!

snario:
Dominator!!!!

Dominator008:
Hmm, I think the actual test is passing, right?

snario:
> Hmm, I think the actual test is passing, right?

There was an issue that @karlfloersch just resolved causing the CI to incorrectly error. Tests are ✅ now!

PR #359, opened 2021-04-05
Have ECDSAContractAccount use max available gas

smartcontracts:
**Description**
Modifies `OVM_ECDSAContractAccount` to pass all available gas to `ovmCALL` or `ovmCREATE`. Gas limit now purely used as a *fee* and does not actually limit the amount of gas that a transaction can use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@ cache-ovm/
 yarn-error.log
 build/
 .DS_STORE
+
+# Coverage output
+coverage/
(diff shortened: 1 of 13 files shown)


karlfloersch (approved):
LGTM!

But @ben-chain can you review?

ben-chain (approved):
Checked out the relevant historic diffs, after considering a bit, LGTM!

PR #361, opened 2021-04-06
Enforce MAX_ROLLUP_TX_SIZE on sequencer tx

maurelian:
`MAX_ROLLUP_TX_SIZE` was enforced on `enqueue()` but not on `appendSequencerBatch()`.


ben-chain (approved):
Looks good to me.  Ran the tests real quick and this does bump per-tx overhead by about 30 gas, worth it!

PR #363, opened 2021-04-06
Prevent reentrancy by disallowing default context values

maurelian:
Fixes: https://github.com/ethereum-optimism/roadmap/issues/803


contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -159,7 +159,9 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         override
         public
     {
-        require(transactionContext.ovmNUMBER == 0, "Only be callable at the start of a transaction");
+        require(transactionContext.ovmNUMBER == 0, "Only callable at the start of a transaction");
+        require(_transaction.blockNumber != 0, "Prevent reentrancy by disallowing default context values");
```
maurelian, line 163: The [issue suggests](https://github.com/ethereum-optimism/roadmap/issues/803#issuecomment-801442414) two options: 

> 1. in run, ban the transaction.blocknumber argument from being passed as its default value, so that you cannot ever get through run in a way which is re-enterable
> 2. add an explicit reentrancy guard variable on run

I initially preferred to add a `nonreentrant` modifier, as it's straightforward and easy to follow. However given that it added  >5000 gas, I decided against this. 

This approach is IMO less readable, but does the trick.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -362,7 +364,7 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
      * Opcodes: Contract Creation *
      ******************************/
 
-    /**
+    /**np
```
ben-chain, line 367: ty ;)

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain, line 163: Agree that this is a fine way to go, although we need to make sure this is handled the same way as an invalid gas limit in `run`.  This is less safe because if you can ever satisfy this require's condition, then you could make proving fraud impossible, since reverts are reserved for `INVALID_STATE_ACCESS`.  Should probably break this and the gaslimit one into a shared internal function.

ben-chain (changes requested):
Looking good, although I think we need to make one minor change in how we prevent this input, see comment below.

One option here is to have an explicit `uint constant DEFAULT_BLOCKNUM_VAL` here which is used both in this check and `_resetContext`?  Seems like that would be a little more robust in future situations where we would want to change the default value, but OTOH there could be other complications and it's a bit ugly.  Thoughts?

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
maurelian: oops.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -67,6 +67,12 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
     uint256 constant NUISANCE_GAS_PER_CONTRACT_BYTE = 100;
     uint256 constant MIN_GAS_FOR_INVALID_STATE_ACCESS = 30000;
 
+    /**************************
+     * Default Context Values *
+     **************************/
+
+    uint256 constant DEFAULT_UINT256 = 0xdefa017defa017defa017defa017defa017defa017defa017defa017defa017d;
```
maurelian, line 74: Previously in #244, I used MAX_UINT, but I feel uneasy about being 'off-by-one' from zero. 

I also wanted a number that can be reused as the default for any of the elements in a  [TransactionContext](https://github.com/ethereum-optimism/contracts/blob/d975c6e524ca9d943ff61f13e2205b0d4c083eaf/contracts/optimistic-ethereum/iOVM/execution/iOVM_ExecutionManager.sol#L47) or [MessageContext](https://github.com/ethereum-optimism/contracts/blob/d975c6e524ca9d943ff61f13e2205b0d4c083eaf/contracts/optimistic-ethereum/iOVM/execution/iOVM_ExecutionManager.sol#L60). 
I like this one, and I don't think it could ever be a valid context number.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1801,7 +1836,7 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
     {
         transactionContext.ovmL1TXORIGIN = address(0);
         transactionContext.ovmTIMESTAMP = 0;
-        transactionContext.ovmNUMBER = 0;
+        transactionContext.ovmNUMBER = DEFAULT_UINT256;
```
maurelian, line 1839: It kind of expands the scope, but I can see an argument fo applying the change to the other values.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -159,7 +164,8 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         override
         public
     {
-        require(transactionContext.ovmNUMBER == 0, "Only be callable at the start of a transaction");
+        require(transactionContext.ovmNUMBER == DEFAULT_UINT256, "Only callable at the start of a transaction");
```
ben-chain, line 167: After discussions with John, realizing that this should return, not revert, so that code contracts cannot maliciously surface a malformed (i.e. not from `_encodeRevertFlag`) revert value.

ben-chain (approved):

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
smartcontracts: It does seem a bit weird for this to be applied to `ovmNUMBER` and not to other values. If we *do* decide to keep the scope limited, we should at least add a comment explaining exactly why we do this for `ovmNUMBER` and not elsewhere.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -159,7 +164,12 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
         override
         public
     {
-        require(transactionContext.ovmNUMBER == 0, "Only callable at the start of a transaction");
+        // Make sure that run() is not re-enterable.  This condition should awlways be satisfied
+        // Once run has been called once, due to the behvaior of _isValidInput().
+        if (transactionContext.ovmNUMBER != DEFAULT_UINT256) {
+            return;
```
smartcontracts, line 170: Hmmm is it ok to return here? If so, why were we reverting before?

ben-chain: I bellieve we explicitly should not have been reverting here, because it would've allowed a malicious caller to revert with malformed data.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1626,6 +1638,33 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
+        // Prevent reentrancy to run():
+        // This check prevents calling run with the default ovmNumber.
+        // Combined with the first check in run():
+        //      if (transactionContext.ovmNUMBER != DEFAULT_UINT256) { return; }
+        // It should be impossible to re-enter since run() returns before any other call frames are created.
+        // Since this value is already being written to storage, we save much gas compared to
+        // using the standard nonReentrant pattern.
+        if (_transaction.blockNumber == DEFAULT_UINT256) return false;
+
+        if (_isValidGasLimit(_transaction.gasLimit, _transaction.l1QueueOrigin) == false) return false;
```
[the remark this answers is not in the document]
ben-chain: FWIW I like keeping it separate; there is a good meaty chunk of logic in there especially with the gas rate limiting stuff.

smartcontracts (approved):

PR #366, opened 2021-04-06
Treat SM address as context

maurelian:
Previously we were resetting the SM address at the end of `run()`. In the case that the `_isValidGasLimit` check fails, this would result in the SM address remaining set, which breaks the invariant that the EM state should remain constant before and after a call to `run()`.


ben-chain (approved):
LGTM!

PR #367, opened 2021-04-07
fix: increase MAX_ROLLUP_TX_SIZE

ben-chain:
**Description**
Bumps the MAX_ROLLUP_TX_SIZE to a more reasonable value so that the new enforcement on sequencer transactions still allows things like large contract deployments.


maurelian (dismissed):
LGTM

maurelian (approved):

PR #369, opened 2021-04-09
More rigorous reverts for unsafe bytecode

ben-chain:
**Description**
This fix allows us to handle unsafe code being returned by initcode much more easily.  While the earlier approach worked in the context of the fraud proof, it required more work to delete the account in geth, as opposed to just reverting a call frame and having it happen naturally.


contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -955,10 +955,24 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-            ? _handleContractCreation(_gasLimit, _data, _contract)
-            : _contract.call{gas: _gasLimit}(_data);
+        bool success; 
+        bytes memory returndata;
+
+        if (_isCreate) {
+            // safeCREATE() is a function which replicates a CREATE message, but uses return values
+            // Which match that of CALL (i.e. bool, bytes).  This allows many security checks to be
+            // to be shared between untrusted call and create call frames.
+            (success, returndata) = address(this).call(
```
smartcontracts, line 965: Can this be replaced with `this.safeCREATE(...)`?

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1026,47 +1040,42 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
         address _address
     )
-        internal
-        returns(
-            bool,
-            bytes memory
-        )
+        external
     {
+        // The only way this should be callable is from within _createContract().
```
smartcontracts, line 1060: Comment is technically wrong in the sense that this check only asserts the caller is the execution manager. More of a nit than anything but it could be confusing to some.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
```diff
@@ -1078,31 +1087,20 @@ contract OVM_ExecutionManager is iOVM_ExecutionManager, Lib_AddressResolver {
...
-            bytes memory revertdata = new bytes(revertDataSize);
-            assembly {
-                returndatacopy(
-                    add(revertdata, 0x20),
-                    0,
-                    revertDataSize
-                )
+            // to be used above in _handleExternalMessage, so we pass the revert data back up unmodified.
+            assembly { 
+                returndatacopy(0,0,returndatasize())
```
smartcontracts, line 1092: Can we leave a comment that we *must* not have any calls in between `Lib_EthUtils.createContract` and this return

test/contracts/OVM/verification/OVM_StateTransitioner.spec.ts
```diff
@@ -312,7 +312,7 @@ describe('OVM_StateTransitioner', () => {
         )
       )
 
-      OVM_StateTransitioner.smodify.set({
+      OVM_StateTransitioner.smodify.put({
```
smartcontracts, line 315: Needs await (+ applies for all instances of `.put` in this file).

smartcontracts (changes requested):
Just a few nits but looks good otherwise. Can approve once addressed.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: I don't think so, because we want access to both of the variables separately like this.

contracts/optimistic-ethereum/OVM/execution/OVM_ExecutionManager.sol
ben-chain: nice

smartcontracts (dismissed):

smartcontracts (approved):

